PR #3, opened 2022-10-10
Update `scripts/daily_snapshot/snapshots_prune.rb`

tyshko5:


lemmih:
Good. Like to be identical to the files in the main repository.

lemmih (approved):

PR #6, opened 2022-10-19
copy some templates and settings from Forest

LesnyRumcajs:
Shamelessly copy some GH templates and goodies (dependabot, codewowners) from the Forest repo.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -3,4 +3,4 @@
 # These owners will be the default owners for everything in
 # the repo. Unless a later match takes precedence,
 # they will be requested for review when someone opens a pull request.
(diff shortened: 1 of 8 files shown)


LesnyRumcajs:
@lemmih It'd be great if you could add @hanabi1224 @creativcoder to the repo. Otherwise the CODEOWNERS file is invalid at the moment.

lemmih:
Maybe we should not add hanabi1224 and creativcoder since they're only here temporarily.

LesnyRumcajs:
> Maybe we should not add hanabi1224 and creativcoder since they're only here temporarily.

Sounds good. In this light, I added @sudo-shashank.

lemmih (approved):

PR #26, opened 2023-01-23
Run forest node in DigitalOcean

samuelarogbonlo:
**Summary of changes**
- New terraform scripts that deploys and run Forest in the cloud. 
- Built to run with just a single command in terraform and ansible

Closes Issue #2422


README.md
```diff
@@ -19,3 +19,32 @@ This repository contains machine-readable specifications for the auxillilary ser
...
+## Architectural Implementation
+
+![Untitled Diagram drawio (8)](https://user-images.githubusercontent.com/47984109/215227510-dac5b8fb-8019-4388-a0e7-d5c432b95d70.png)
+
+The flow goes from Terraform for provisioning of the servers and Anisble to run all neccessary installations including forest.
+
+## Requirements 
+- RAM: 32GB
+- VCPU: 8
+- Startup Disk Size: 100 GB
```
LesnyRumcajs, line 34: For startup, we need at least 200G, +100G to download the snapshot and another +100G to fill the DB. After this process, we can delete the snapshot.

README.md
```diff
@@ -19,3 +19,32 @@ This repository contains machine-readable specifications for the auxillilary ser
...
+- Expected Total Disk Size: > 500 GB
+
+N/B: It's worth to note that some of the naming conventions can be changed to suit your deployment needs. 
+
+To test out the implementation, just access the server with appropiate `ssh` details in this manner `ssh -i ~/id_rsa chainsafe@ip_address`.
+
+Also, out of curiosity, to initiate the infrastructure, run the following:
+- Ensure to add all neccessary enviroment variables for AECCESS_ID and SECRET_KEY into the local machine via the `echo` command. This is needed for the backend and state configurations.
```
LesnyRumcajs, line 42: `ACCESS_ID`? Where do you get those variables? How do you add them via `echo` command? Assume the reader is not proficient and definitely assume if you make a typo (s)he will copy-paste it.

README.md
```diff
@@ -19,3 +19,32 @@ This repository contains machine-readable specifications for the auxillilary ser
...
+- Run `make apply` in the terraform directory.
+- Run `ansible-playbook forest-docker-run.yaml` in the ansible directory.
+
+## Collaborators
+- [YOUR NAME HERE] - Feel free to contribute to the codebase by resolving any open issues, refactoring, adding new features, writing test cases or any other way to make the project better and helpful to the community. Feel free to fork and send pull requests.
```
LesnyRumcajs, line 47: What's the purpose of the `[YOUR NAME HERE]`?

README.md
```diff
@@ -19,3 +19,32 @@ This repository contains machine-readable specifications for the auxillilary ser
...
+
+## Questions
+Feel free to reach out to the BlockOps team for more details on how to interact with the infrastructure if the need arises while in deployment.
```
LesnyRumcajs, line 50: How to reach the BlockOps team?

ansible/ansible.cfg
```diff
@@ -0,0 +1,8 @@
+[defaults]
+host_key_checking = false
+inventory = hosts
+
+remote_user = chainsafe
+private_key_file = ~/sammy
```
LesnyRumcajs, line 6: What is this file? Is it yours?

ansible/roles/docker/defaults/main.yml
```diff
@@ -0,0 +1,17 @@
+---
+# defaults file for role/docker
+
+install_docker: "{{ install_docker }}"
+docker_edition: "ce"
+docker_pkg: "docker-{{ docker_edition }}"
+
+docker_apt_ignore_key_error: True
+docker_apt_key: "9DC858229FC7DD38854AE2D88D81803C0EBFCD88"
```
LesnyRumcajs, line 9: what is this key?

ansible/roles/docker/tasks/main.yml
```diff
@@ -0,0 +1,120 @@
...
+- name: Create a login user
+  user:
+    name: "{{ user }}"
+    state: present
+
+- name: Add public key to authorized_keys
+  authorized_key:
+    user: "{{ user }}"
+    state: present
+    key: "{{ lookup('file', '~/sammy.pub') }}"
```
LesnyRumcajs, line 15: what is this file?

ansible/roles/docker/tasks/main.yml
```diff
@@ -0,0 +1,120 @@
...
+  shell: sudo service ssh restart 
+
+- name: Allow specific users to log in
+  ansible.builtin.lineinfile:
+    dest: /etc/ssh/sshd_config
+    regexp: '^AllowUsers'
+    line: 'AllowUsers {{ user }}'
+    state: present
+
+- name: Add {{ user }} to sudoers file
```
LesnyRumcajs, line 35: why do we need it in sudoers?

ansible/roles/docker/tasks/main.yml
```diff
@@ -0,0 +1,120 @@
+---
```
LesnyRumcajs, line 1: this should be rather `roles/service-user`, no?

ansible/roles/docker/tasks/main.yml
```diff
@@ -0,0 +1,120 @@
...
+    - software-properties-common
+
+- name: Add Docker's official GPG key
+  apt_key:
+    url: https://download.docker.com/linux/ubuntu/gpg
+    id: "{{ docker_apt_key }}"
+    state: present
+  ignore_errors: True 
+
+- name: Add Docker repository
```
LesnyRumcajs, line 69: Isn't `docker` available out of the box?

terraform/backend.tf
```diff
@@ -0,0 +1,10 @@
+terraform {
+  backend "s3" {
+    bucket = "forest-test-state"
+    key = "terraform.tfstate"
+    region = "us-west-1"
+    endpoint = "nyc3.digitaloceanspaces.com"
+    skip_credentials_validation = true
+    skip_metadata_api_check = true
```
LesnyRumcajs, line 8: Why?

terraform/backend.tf
```diff
@@ -0,0 +1,10 @@
+terraform {
+  backend "s3" {
+    bucket = "forest-test-state"
+    key = "terraform.tfstate"
+    region = "us-west-1"
+    endpoint = "nyc3.digitaloceanspaces.com"
```
LesnyRumcajs, line 6: Why? This should be customisable with variables. Plus, most of our infra runs on FRA1, let's stick to this region as the default. `NYC` is far for most of the team.

terraform/backend.tf
```diff
@@ -0,0 +1,10 @@
+terraform {
+  backend "s3" {
```
LesnyRumcajs, line 2: Why do we need `s3`?

terraform/main.tf
```diff
@@ -0,0 +1,115 @@
...
+# resource "digitalocean_ssh_key" "new-key" {
+#   name       = "new-key"
+#   public_key = file("/Users/mac/.ssh/id_rsa.pub")
+# }
+
+# resource "digitalocean_spaces_bucket" "spaces-name" {
+#   name   = "spaces-name"
+#   region = "nyc3"
+# }
+
```
LesnyRumcajs, line 25: Why do we need this?

terraform/main.tf
```diff
@@ -0,0 +1,115 @@
...
+resource "digitalocean_volume" "forest-volum" {
+  region                  = "nyc3"
```
LesnyRumcajs, line 27: the region should be configurable in a single place

terraform/main.tf
```diff
@@ -0,0 +1,115 @@
...
+#   name       = "new-key"
+#   public_key = file("/Users/mac/.ssh/id_rsa.pub")
+# }
+
+# resource "digitalocean_spaces_bucket" "spaces-name" {
+#   name   = "spaces-name"
+#   region = "nyc3"
+# }
+
+resource "digitalocean_volume" "forest-volum" {
```
LesnyRumcajs, line 26: It should be rather `forest-volume`, no? Or better, something parameterizable.

terraform/main.tf
```diff
@@ -0,0 +1,115 @@
...
+  region                  = "nyc3"
+  name                    = "forest-volum"
+  size                    = 600
+  initial_filesystem_type = "ext4"
+  description             = "forest storage volume"
+}
+
+resource "digitalocean_droplet" "forest-samuel" {
```
LesnyRumcajs, line 34: Your name should not be present in those files

terraform/main.tf
```diff
@@ -0,0 +1,115 @@
...
+  image  = var.image
+  name   = var.name
+  region = var.region
+  size   = var.size
+  backups = var.backups
+  ssh_keys = [var.sam_ssh_key_fingerprint, var.guillaume_ssh_key_fingerprint, var.hubert_ssh_key_fingerprint, var.david_ssh_key_fingerprint]
```
LesnyRumcajs, line 40: Can we get this from an optional configuration file?

LesnyRumcajs (changes requested):
Thanks for the effort, we're going in the right direction. Some comments so far:

* Please delete and ignore `.DS_Store` files. As far as I know, those are some MacOS artefacts.
* Please justify with a comment whenever you are skipping some credential checks,
* `forest-samuel` should not be present in the files. It's okay for testing, but a prod-like node should be completely agnostic in this manner.
* the entire thing should be more configurable

README.md
```diff
@@ -19,3 +19,32 @@ This repository contains machine-readable specifications for the auxillilary ser
 - [ ] Exhaustive RPC testing for Forest docker image.
 - [ ] Sync testing for PRs.
 - [ ] Export testing for PRs.
+
+## Running Forest In DigitalOcean
+
+## Architectural Implementation
+
+![Untitled Diagram drawio (8)](https://user-images.githubusercontent.com/47984109/215227510-dac5b8fb-8019-4388-a0e7-d5c432b95d70.png)
```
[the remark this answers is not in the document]
samuelarogbonlo: Are you referring to the name "Architectural Implementation"?

README.md
samuelarogbonlo: On it too.

ansible/ansible.cfg
samuelarogbonlo: Changing it now to `id_rsa`

ansible/roles/docker/defaults/main.yml
samuelarogbonlo: It's a docker apt-key, it's on docker's documentation

ansible/roles/docker/tasks/main.yml
samuelarogbonlo: Changing it now to the default public key file-name

ansible/roles/docker/tasks/main.yml
samuelarogbonlo: This is to give the new user ability to still run root commands if the critical need arises but it's a very unlikely scenario

ansible/roles/docker/tasks/main.yml
samuelarogbonlo: It's not a bad idea tho, but I just used `main.yaml` since we do not have multiple files in the directory

ansible/roles/docker/tasks/main.yml
samuelarogbonlo: No it is not

terraform/main.tf
samuelarogbonlo: sure

terraform/main.tf
samuelarogbonlo: The thing about this is that the `terraform.tfvars` file has the configurations. So from the gitignore content, any user should know there needs to be a `.tfvars` file with the correct values. However, I'd add a note to the docs to let users know the details.

terraform/main.tf
samuelarogbonlo: It's a template to create the key and spaces; but I'd make it reconfigurable now and uncomment

terraform/main.tf
samuelarogbonlo: Sure

terraform/main.tf
samuelarogbonlo: Sure

README.md
samuelarogbonlo: I'd change the comment; we should rather go with letting the fellow raise an issue or something

terraform/backend.tf
samuelarogbonlo: Oh okay. FRA1 it is then.

terraform/backend.tf
samuelarogbonlo: We need a backend configuration to manage state files or else the users will either have to host the state locally or b at risk of losing the state of terraform implementation (in simple terms, it means, everytime terraform is initaited, it does not continue from the last deployment, it destroys available resources and create new ones).

terraform/backend.tf
samuelarogbonlo: This can be useful in cases where Terraform cannot automatically determine the validity of the credentials or the metadata API availability. However, the user still wants to proceed with the Terraform operation. By skipping the validation checks, Terraform assumes that the provided credentials are valid and the metadata API is available and will proceed with the process.

README.md
```diff
@@ -19,3 +19,34 @@ This repository contains machine-readable specifications for the auxillilary ser
...
+- VCPU: 8
+- Startup Disk Size: 200 GB
+- Expected Total Disk Size: > 500 GB
+
+N/B: It's worth to note that most of the naming conventions can be changed to suit your deployment needs especially the names for the resource blocks. In this case; `forest-volume`, `new-key-name`,`spaces-name`,`forest-volume`, `forest`, `forest-firewalls-test` and `hosts`.
+
+To test out the implementation, just access the server with appropiate `ssh` details in this manner `ssh -i ~/.ssh/id_rsa chainsafe@ip_address`. However, if you intend to use another `user` other than `chainsafe`, this can be switched in the Ansible main playbook, just avoid running forest as root.
+
+In order ro implement the infrastructure, run the following:
+- Set Up s3cmd 2.x with DigitalOcean Spaces; you can check [here](https://docs.digitalocean.com/products/spaces/reference/s3cmd/) on more details. This will require `ACCESS_TOKEN` and `SECRET_KEY` but it can be automatically generated from the DigitalOcean console through the Applications & API section.
```
LesnyRumcajs, line 42:
```suggestion
- Set-up `s3cmd` 2.x with DigitalOcean Spaces; you can check [here](https://docs.digitalocean.com/products/spaces/reference/s3cmd/) on more details. This will require `ACCESS_TOKEN` and `SECRET_KEY`; it can be automatically generated from the DigitalOcean console through the Applications & API section.  
```

README.md
```diff
@@ -19,3 +19,34 @@ This repository contains machine-readable specifications for the auxillilary ser
...
+- Startup Disk Size: 200 GB
+- Expected Total Disk Size: > 500 GB
+
+N/B: It's worth to note that most of the naming conventions can be changed to suit your deployment needs especially the names for the resource blocks. In this case; `forest-volume`, `new-key-name`,`spaces-name`,`forest-volume`, `forest`, `forest-firewalls-test` and `hosts`.
+
+To test out the implementation, just access the server with appropiate `ssh` details in this manner `ssh -i ~/.ssh/id_rsa chainsafe@ip_address`. However, if you intend to use another `user` other than `chainsafe`, this can be switched in the Ansible main playbook, just avoid running forest as root.
+
+In order ro implement the infrastructure, run the following:
+- Set Up s3cmd 2.x with DigitalOcean Spaces; you can check [here](https://docs.digitalocean.com/products/spaces/reference/s3cmd/) on more details. This will require `ACCESS_TOKEN` and `SECRET_KEY` but it can be automatically generated from the DigitalOcean console through the Applications & API section.  
+- Ensure to add all neccessary enviroment variables by running `echo $SPACES_ACCESS_TOKEN` and `echo $SPACES_SECRET_KEY` while using the values generated in the previous step.
```
LesnyRumcajs, line 43:
```suggestion
- add all necessary environment variables by running `echo $SPACES_ACCESS_TOKEN` and `echo $SPACES_SECRET_KEY` while using the values generated in the previous step.
```

LesnyRumcajs, line 43: Printing the variables' values to the console doesn't add them anywhere, no?

README.md
```diff
@@ -19,3 +19,34 @@ This repository contains machine-readable specifications for the auxillilary ser
...
+- Create a `*.tfvars` file to use the neccessary variable values for terraform running success. Also, be aware that the `keys_name` must be unique.
```
LesnyRumcajs, line 44: What are the necessary values? Let's have a template.
```suggestion
- Create a `*.tfvars` file to use the necessary variable values for terraform running success. Also, be aware that the `keys_name` must be unique.  
```

LesnyRumcajs (changes requested):
I'll review the rest later; got a meeting to attend. It'd be fantastic if you ran the README through Grammarly or another typo/grammar checker. Also, I still see the `.DS_Store` files. Please remove them.

A general thing that came to my mind - this repository is supposed to hold all infrastructure code, so it'd be best to put all the changes related to the node under the `forest_node` directory. You can then link to it from the main README.

README.md
samuelarogbonlo: Nope it does not

README.md
samuelarogbonlo: I think this is not necessary. Anybody that sees the `variables.tf` files can tell what variable values need to go to the `.tfvars` file cos the variables are already declared.

samuelarogbonlo:
> I'll review the rest later; got a meeting to attend. It'd be fantastic if you ran the README through Grammarly or another typo/grammar checker. Also, I still see the `.DS_Store` files. Please remove them.
> 
> A general thing that came to my mind - this repository is supposed to hold all infrastructure code, so it'd be best to put all the changes related to the node under the `forest_node` directory. You can then link to it from the main README.

I do not think the `.DS_Store` still exists because I don't have it in my working branch. Also, I have moved the main `forest_node` details for installation to the right directory. However, I do not think there's a need to link it from the main readme; for sure, anyone who accesses the `forest_node` directory will see all the necessary details.

README.md
LesnyRumcajs: So what do you achieve by printing them?

forest-node/ansible/ansible.cfg
```diff
@@ -0,0 +1,8 @@
+[defaults]
+host_key_checking = false
+inventory = hosts
+
+remote_user = chainsafe
+private_key_file = ~/id_rsa
```
LesnyRumcajs, line 6: What is this private key exactly? Why does ansible need it? Why is it `id_rsa`? On my machine, it's not the case.

forest-node/ansible/ansible.cfg
```diff
@@ -0,0 +1,8 @@
...
+
+allow_world_readable_tmpfiles=true
```
LesnyRumcajs, line 8: why?

forest-node/ansible/roles/docker/tasks/main.yml
```diff
@@ -0,0 +1,120 @@
...
+  shell: sudo service ssh restart 
+
+- name: Allow specific users to log in
+  ansible.builtin.lineinfile:
+    dest: /etc/ssh/sshd_config
+    regexp: '^AllowUsers'
+    line: 'AllowUsers {{ user }}'
+    state: present
+
+- name: Add {{ user }} to sudoers file
```
LesnyRumcajs, line 35: It's practically a root user. Let's allow logging as' root' if we want to have a root user and leave the service user limited.

forest-node/ansible/roles/docker/tasks/main.yml
```diff
@@ -0,0 +1,120 @@
...
+    state: present
+  with_items:
+    - acl
+    - python3-pip 
+    - apt-transport-https
+    - ca-certificates
+    - curl
+    - software-properties-common
+
+- name: Add Docker's official GPG key
```
LesnyRumcajs, line 62: This all can be omitted by using a proper image I linked earlier.

LesnyRumcajs (changes requested):
We're getting there. I believe using https://marketplace.digitalocean.com/apps/docker will significantly reduce the docker boilerplate, so it'd be great if you could try it out.

README.md
samuelarogbonlo: Oh when you said printing, I thought of something different. Anyways, this stores as part of then `env` values in the local machine while attempting to run the commands.

forest-node/ansible/ansible.cfg
samuelarogbonlo: The private key is needed for ansible to successfully `ssh` into the server. The `id_rsa` is the conventional name given to private keys when they are created but for sure it is different in your machine because you put a custom name upon creation. 

If you use the terminal suggestion, it will be `id_rsa` and `id_rsa.pub`

forest-node/ansible/roles/docker/tasks/main.yml
samuelarogbonlo: Got it. I'd allow root user login and limit the service user.

forest-node/ansible/ansible.cfg
```diff
@@ -0,0 +1,8 @@
+[defaults]
+host_key_checking = false
```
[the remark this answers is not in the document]
samuelarogbonlo: Yes, the host-checking might be cool but we want the easiest way possible for anybody to run this. In the case of this check, assuming a host is reinstalled and has a different key in ‘known_hosts’, there will be an error message until it is corrected (from experience). 

So If a new host is not in ‘known_hosts’, the control node may prompt for the confirmation of the key, which results in some form of manual/interactive intervention. So to avoid unevenness in the behaviors, then we disable host checking.

forest-node/ansible/ansible.cfg
samuelarogbonlo: This is more like custom best practice, especially when running a new user (in our case, one with limited service privileges). So this command ensures that the temporary files created on the server are world readable and will display a level of warning instead of failing the running task.

It's more about general good practice with new users

forest-node/ansible/roles/docker/defaults/main.yml
```diff
@@ -0,0 +1,17 @@
+---
```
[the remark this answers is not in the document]
samuelarogbonlo: Whew! What this means is I should use terraform to initialize a `server using Docker image` and then run forest on it. Not a bad idea but I do not think the way we are running it now is a problem because this is a straight lift from docker docs. 

However, I'd try this idea, I just have to be sure it can offer enough server, CPU and RAM expectations

forest-node/ansible/roles/docker/handlers/main.yml
```diff
@@ -0,0 +1,4 @@
+---
```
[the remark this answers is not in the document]
samuelarogbonlo: There are some tasks that may require docker restart in the future (maybe) so we make docker flexible for restart at any time. Generally, handlers are used to maybe start, reload, restart, and stop running services. 

If the playbook involves changing configuration files, there is a high chance that we might need to restart a service so that the changes take effect. We can add maybe a trigger word `run`, `notify`, or something in the tasks that need the service restart before making the tasks successful.  

So once again, this very handler is about best practice.

forest-node/ansible/ansible.cfg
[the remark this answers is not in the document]
samuelarogbonlo: Also, I must add that this increases speed of configurations with ansible

README.md
LesnyRumcajs: Sorry, I don't get what you achieve by echoing them. What happens if you don't do it?

forest-node/ansible/ansible.cfg
LesnyRumcajs: So the user will get a random error from Ansible in case he doesn't use the default name. It's a bit harsh. I recommend putting it as a variable. Ultimately, when I run this on my machine I shouldn't have any changes in the tracked source files. Do you agree?

forest-node/ansible/ansible.cfg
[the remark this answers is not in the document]
LesnyRumcajs: Thanks for explaining. It's probably okay to leave it like this.

README.md
samuelarogbonlo: If you don't do it then it'll be difficult to setup the backend for the terraform config (which means no spaces and storage of state files)

forest-node/ansible/ansible.cfg
samuelarogbonlo: I don't think I agree because in the requirements step, it is necessary to `ssh keygen` which means before even attempting to run the playbook, the files should have been created if the user follows the docs

forest-node/ansible/roles/docker/tasks/main.yml
samuelarogbonlo: sure

forest-node/README.md
```diff
@@ -0,0 +1,35 @@
+
+## Running Forest In DigitalOcean
+
+## Forest IAC Architectue
+
+![Forest Cloud Infrastructure ](https://user-images.githubusercontent.com/47984109/216006502-eca661d3-2ef8-4c75-aa7a-1740c25abb44.png)
+
+## Requirements 
+- RAM: 32GB
+- VCPU: 8
```
lemmih, line 10: 16GB with 2 cores should suffice but I guess this is something we can hone later.

forest-node/ansible/roles/forest/tasks/main.yml
```diff
@@ -0,0 +1,62 @@
...
+- name: Create a login user
+  user:
+    name: "{{ user }}"
+    state: present
+
+- name: Add public key to authorized_keys
+  authorized_key:
+    user: "{{ user }}"
+    state: present
+    key: "{{ lookup('file', '~/sammy.pub') }}"
```
lemmih, line 15: What does this do?

forest-node/terraform/backend.tf
```diff
@@ -0,0 +1,10 @@
+terraform {
+  backend "s3" {
+    bucket = "forest-test-state"
```
lemmih, line 3: This information should be mentioned in the README

forest-node/ansible/roles/forest/tasks/main.yml
```diff
@@ -0,0 +1,62 @@
...
+- name: Create a login user
+  user:
+    name: "{{ user }}"
+    state: present
+
+- name: Add public key to authorized_keys
+  authorized_key:
+    user: "{{ user }}"
+    state: present
+    key: "{{ lookup('file', '~/.ssh/id_rsa.pub') }}"    # assuming the name of the public key file is `id_rsa.pub`
```
lemmih, line 15: I don't have my public key at `~/.ssh/id_rsa.pub`. How about managing the keys through Digital Ocean?

forest-node/ansible/roles/forest/tasks/main.yml
```diff
@@ -0,0 +1,62 @@
...
+
+- name: Allow specific users to log in
+  ansible.builtin.lineinfile:
+    dest: /etc/ssh/sshd_config
+    regexp: '^AllowUsers'
+    line: 'AllowUsers {{ user }}'
+    state: present
```
lemmih, line 22: Is this necessary?

forest-node/ansible/roles/forest/tasks/main.yml
```diff
@@ -0,0 +1,62 @@
+---
+# tasks file for role/forest
+- name: Upgrade all apt packages
+  apt: upgrade=dist force_apt_get=yes
```
lemmih, line 4: Is it necessary to upgrade the packages?

forest-node/ansible/ansible.cfg
```diff
@@ -0,0 +1,8 @@
...
+inventory = hosts
+
+remote_user = root
+private_key_file = ~/.ssh/id_rsa    # assuming the name of the private key file is `id_rsa`
```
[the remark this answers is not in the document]
samuelarogbonlo: No, I think this is an over-engineering process, we need ansible to access the key file in the user's directory. Every default key creation process stores in this directory, but if the keys are located elsewhere for any reason, the user needs to update the directory there as the case applies.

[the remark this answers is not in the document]
lemmih: Why not use an authentication agent instead? Pretty sure ansible supports that natively.

forest-node/ansible/roles/forest/tasks/main.yml
samuelarogbonlo: Yes actually - this is synonymous with `sudo apt-get update`

forest-node/ansible/roles/forest/tasks/main.yml
```diff
@@ -0,0 +1,62 @@
...
+    name: '{{ user }}'
+    append: true
+    shell: /bin/bash
+    system: true
+    create_home: false
+
+- name: update all packages 
+  become: true
+  shell: 
+    cmd: sudo apt update
```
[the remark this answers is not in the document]
samuelarogbonlo: It is required to update all the packages again before running forest - I had errors while testing.

forest-node/ansible/roles/forest/tasks/main.yml
lemmih: Why is it necessary?

forest-node/ansible/roles/forest/tasks/main.yml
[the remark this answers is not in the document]
lemmih: What errors? If this step is truly necessary then we might need to add it to the README of Forest.

forest-node/ansible/roles/forest/tasks/main.yml
samuelarogbonlo: Just a best practice is to allow just a specific user to interact with the server - basic security play

forest-node/ansible/roles/forest/tasks/main.yml
samuelarogbonlo: You just have to change the directory to where your public key exists.

lemmih: Why can't we either put the key in a configuration file or (better yet) just manage the keys through Digital Ocean?

forest-node/ansible/roles/forest/tasks/main.yml
lemmih: Unless I'm mistaken, the Digital Ocean image we're using already allows users to connect via SSH. Are you saying this is not the case?

forest-node/terraform/backend.tf
samuelarogbonlo: This is in the README

forest-node/README.md
samuelarogbonlo: Yes sure

lemmih: Let's update the README to match the specs of the droplet we're using by default.

forest-node/ansible/roles/forest/tasks/main.yml
samuelarogbonlo: It is still `ssh` being used here; we are just using a new user to log in to the server and blocking root access.

forest-node/ansible/roles/forest/tasks/main.yml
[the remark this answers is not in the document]
samuelarogbonlo: I cannot recall the error specifically. But by and large, there is no harm in updating packages after running certain processes plus definitely, we want to keep all the packages on the host up to date.

forest-node/ansible/roles/forest/tasks/main.yml
samuelarogbonlo: This is a server just created and normal practice is to update which simply makes sure our list of packages from all repositories and PPAs is up to date. If we do not run this command, we could be getting older versions of various packages we install, or worse, dependency issues. 

This is just good practice when interacting with Linux servers

forest-node/ansible/ansible.cfg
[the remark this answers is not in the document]
samuelarogbonlo: Auth agent? I am not exactly clear on that - do you mean the likes of OKTA, teleport and co?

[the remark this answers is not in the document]
lemmih: Like [ssh-agent](https://www.ssh.com/academy/ssh/agent). I'm pretty sure you're already using it. For example, when you push commits to this repository, you don't have to unlock your private every time. The ssh agent takes care of it for you.

forest-node/ansible/roles/forest/tasks/main.yml
lemmih: Whether this is good practice is highly debated. Nix, for example, has the complete opposite point of view and I'm quite partial to this way of seeing things. If there's a security vulnerability in the image we're using then we'll update the image rather updating the software in-place.

forest-node/ansible/roles/forest/tasks/main.yml
samuelarogbonlo: For sure, yes and allow only the specified user

forest-node/ansible/roles/forest/tasks/main.yml
lemmih: But we already created a set of authorized keys when we created the droplet, right? Can't we reuse those authorized keys?

forest-node/README.md
samuelarogbonlo: Sure - done.

forest-node/ansible/roles/forest/tasks/main.yml
samuelarogbonlo: Well, afaik, I am not aware of any vulnerability - but I'd try the run the play now w/o any of the package update process and see how it goes

forest-node/ansible/roles/forest/tasks/main.yml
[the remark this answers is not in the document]
samuelarogbonlo: Got it

forest-node/ansible/roles/forest/tasks/main.yml
samuelarogbonlo: Just tested this and it works w/o updating packages so i'd push the changes

forest-node/ansible/roles/forest/tasks/main.yml
samuelarogbonlo: Yes, we are using the same keys but in this case, we are adding it for the new user to log in via its own authorized-keys folder. While root has its own, the new user needs its own as well

lemmih: Can't we copy the authorized keys from the root user?

samuelarogbonlo: Erm, well, this is a possibility (I have an imagination of the playbook in my head already) but can I understand what the issue is here exactly? Is there a problem or disadvantage with copying the keys from the local machine 

Meanwhile, I'd cook up a play now and try if that works as well.

lemmih: The only problem is that `~/.ssh/id_rsa.pub` might not be where users keep their public key. And we're already specifying the public key using a fingerprint. It would simplify the process if we only used the fingerprint.

forest-node/ansible/ansible.cfg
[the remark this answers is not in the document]
samuelarogbonlo: I am still looking at ssh-agent tho, but it requires some manual process before ansible takes over which I am most definitely trying to avoid cos' I am moving from terraform to ansible at once. 

For me to set up ssh-agent, I need to access the server which will still require Ansible in our case and the whole purpose of `ansible.cfg` is to even initiate the login and connection process between Ansible and the server. 

I am still trying to dig but at first glance, this agent may not be able to do this perfectly.

forest-node/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+![Forest Cloud Infrastructure ](https://user-images.githubusercontent.com/47984109/219903795-77c306b8-a70b-4f32-8d7d-3c1a39e52186.jpg)
+
+## Requirements  
+- RAM: 16GB
+- VCPU: 2
+- Startup Disk Size: 200 GB
+- Expected Total Disk Size: > 500 GB
+- SSH Key should be created locally using `ssh keygen` and then added into digitalocean console where the fingerprint can be generated and added as a variable while creating the droplet.
+- Install terraform and ansible.
```
lemmih, line 14: Could link to https://developer.hashicorp.com/terraform/downloads and https://docs.ansible.com/ansible/latest/installation_guide/intro_installation.html

forest-node/terraform/Makefile
```diff
@@ -0,0 +1,11 @@
+init:
+	@terraform init"
```
lemmih, line 2: Is this a typo? I can't seem to make it work.

lemmih: After removing the `"`, I get this error:
```
Initializing the backend...
╷
│ Error: error configuring S3 Backend: no valid credential sources for S3 Backend found.
│ 
│ Please see https://www.terraform.io/docs/language/settings/backends/s3.html
│ for more information about providing credentials.
│ 
│ Error: NoCredentialProviders: no valid providers in chain. Deprecated.
│ 	For verbose messaging see aws.Config.CredentialsChainVerboseErrors
│ 
│ 
│ 
```

I've already configured `s3cmd`:
```
$ s3cmd ls
2022-07-12 13:29  s3://forest-snapshots
2021-11-26 19:23  s3://lodestar-misc
```

forest-node/README.md
samuelarogbonlo: Got it.

forest-node/terraform/Makefile
samuelarogbonlo: I've made some changes in regards to `terraform init` but have you already created the space bucket and attached the name to the `backend.tf` file ?

LesnyRumcajs (changes requested):

lemmih (changes requested):
Looks pretty good. A few minor comments:

- `terraform.tfvars` contains four important variables and six variables that should almost never change. How about we move the important variables to the top? And is it possible to add comments to the file?
- You often omit a trailing newline. This shows up as a warning in GitHub. Mind adding the newlines?
- How come `digitalocean_token` is defined in the variable file but the AWS secrets are not? Wouldn't it be easier to keep all the secrets together in the same place?
- Can we figure out a way to not define the firewall name?
- I'm still confused about the `create docker user` ansible step. Could you explain it to me?

Issue #36, opened 2023-02-28
Add Docker linter to the CI


LesnyRumcajs:
**Issue summary**

Add Docker linter to the CI
https://github.com/hadolint/hadolint

Double-check that all Dockerfiles (and potentially docker-compose files) are verified. 

**Other information and links**

Issue #34, opened 2023-02-20
Lock dependencies for Ruby scripts


LesnyRumcajs:
**Issue summary**

As per @hanabi1224 suggestion, the dependencies of Ruby scripts are not bound to specific versions, which can introduce regressions out of nowhere. It didn't happen yet, but better safe than sorry.

This can be resolved using gem files, specifically having `Gemfile.lock` tracked.

**Other information and links**

PR #47, opened 2023-03-13
hadolint ci changes

samuelarogbonlo:
**Summary of changes**
-  Setting up CI workflow to check docker files

This is in reference to issue #36

### DIFF
--- a/.github/workflows/hadolint-ci.yml
+++ b/.github/workflows/hadolint-ci.yml
@@ -1,0 +1,24 @@
+name: 'Hadolint Linting'
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+jobs:
+  hadolint:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Install hadolint
+        run: sudo wget -O /bin/hadolint https://github.com/hadolint/hadolint/releases/download/v2.7.0/hadolint-Linux-x86_64 && sudo chmod +x /bin/hadolint
+
+      - name: Lint Dockerfiles
+        run: hadolint **/Dockerfile**
+
+      - name: Lint Docker Compose files
+        run: hadolint **/docker-compose**
+


.github/workflows/hadolint-ci.yml
```diff
@@ -0,0 +1,20 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Install hadolint
+        run: sudo wget -O /bin/hadolint https://github.com/hadolint/hadolint/releases/download/v2.7.0/hadolint-Linux-x86_64 && sudo chmod +x /bin/hadolint
+
+      - name: Lint Dockerfiles
+        run: hadolint **/Dockerfile**
```
LesnyRumcajs, line 20: I don't think there's a need for such manual installation. We can use the hadolint container.
```suggestion
  lint:
    runs-on: ubuntu-latest
    container: hadolint/hadolint:latest-debian
    steps:
    - uses: actions/checkout@v2
    - name: hadolint
      run: hadolint **/*Dockerfile*
```

the hadolint action should also work https://github.com/hadolint/hadolint-action

see the draft PR https://github.com/ChainSafe/forest/pull/2615/files (you may need to enable recursive search)

LesnyRumcajs (dismissed):

LesnyRumcajs:
@samuelarogbonlo you also need to fix the issues reported by the linter in this PR. No green checkmark, no merge. :)

scripts/Dockerfile-sync-check
```diff
@@ -1,12 +1,28 @@
...
+RUN gem install \
+    docker-api \
+    slack-ruby-client \
+    activesupport
 
+# Set a working directory and copy files
+WORKDIR /usr/src/app
 COPY ruby_common ruby_common
-COPY sync_check/* ./
+COPY daily_snapshot/*.rb daily_snapshot/upload_snapshot.sh daily_snapshot/run_service.sh ./
```
lemmih, line 25: This does not look right. You copy the daily snapshot files instead of the sync check files? That appears to be a mistake.

.github/workflows/hadolint-ci.yml
```diff
@@ -0,0 +1,23 @@
...
+    steps:
+    - uses: actions/checkout@v2
+
+    - uses: hadolint/hadolint-action@v3.1.0
+      with:
+        dockerfile: "Dockerfile*"
+        recursive: true
+        # Ignore https://github.com/hadolint/hadolint/wiki/DL3041
+        # Ignore https://github.com/hadolint/hadolint/wiki/DL3028
+        # Ignore https://github.com/hadolint/hadolint/wiki/DL3050
```
lemmih, line 21: Could you write a comment describing why we have these lint exceptions?

lemmih (changes requested):

.github/workflows/hadolint-ci.yml
samuelarogbonlo: Would you like this comment in the code or here? However, the wiki links in the comments detail why we had to ignore those errors.

lemmih: I don't see anything in the wiki about why we are ignoring these lints.

scripts/Dockerfile-sync-check
samuelarogbonlo: Yeah sure - I just corrected it.

scripts/Dockerfile-sync-check
```diff
@@ -1,12 +1,27 @@
...
+    && dnf clean all
 
-WORKDIR /chainsafe
+# Gem install dependencies
+RUN gem install \
+    docker-api \
+    slack-ruby-client \
+    sys-filesystem
 
+# Set a working directory and copy files
```
LesnyRumcajs, line 21: Are those comments really helpful?

scripts/sync_check/Dockerfile-tester
```diff
@@ -1,6 +1,15 @@
...
+    ruby-devel \
+    make \
+    gcc \
+    && dnf clean all
 
 # Install required Ruby packages
-RUN gem install docker-api slack-ruby-client
+RUN gem install \
+    docker-api \
+    slack-ruby-client
```
LesnyRumcajs, line 15: Same. Alternatively we do it in a separate issue https://github.com/ChainSafe/forest-iac/issues/34

scripts/Dockerfile-sync-check
samuelarogbonlo: Well, I just felt it'd be good to make any reader understand the operations of each line.

scripts/Dockerfile-sync-check
```diff
@@ -1,12 +1,27 @@
...
+    make \
+    gcc \
+    && dnf clean all
 
-WORKDIR /chainsafe
+# Gem install dependencies
+RUN gem install \
+    docker-api \
+    slack-ruby-client \
+    sys-filesystem
```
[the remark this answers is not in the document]
samuelarogbonlo: Hmm, I'd have to look into this

.github/workflows/hadolint-ci.yml
samuelarogbonlo: Since we are using `gem` and `dnf` to install the packages, for the error to be fixed, we need to hardcode the preferred version number to each package. But in order to always have the updated version, we avoid hardcoding the numbers and ignore the error (this is for the first two ignores)

Also for the last ignore process, we cannot add a label at the beginning of the docker file cos that feature has been deprecated for Security reason, however since the repo is open-source, we can add label and run the ignore.

samuelarogbonlo: It does not matter if we use `gem` and `dnf`, even `apt` has the same issues. But this is AFAIK tho, you can check the wiki for more information on my context.

samuelarogbonlo: Well, following @LesnyRumcajs review, I'd have to pin the versions and remove the ignore so going ahead with that.

scripts/Dockerfile-sync-check
LesnyRumcajs: We should assume that all readers of a Dockerfile should know those basic keywords. Otherwise, we'll end up in a grotesque situation like this:
```
# change to the parent directory
cd ..
# print "finished" to the standard output
echo "finished"
# exit the script with code zero, which indicates a success
exit 0
```

I'm okay with documenting non-trivial or domain-specific code. But the comments should explain things, not reword the code.

For example:
```
# Gem install dependencies
RUN gem install <dependencies>
```
You are installing some dependencies. It's obvious. What we can add more to it, for those who never did any Ruby (a reasonable assumption that most people didn't)
```
# Install dependencies for the Ruby script
RUN gem install <dependencies>
```
This gives a link between `gem` and Ruby to the reader.

.github/workflows/hadolint-ci.yml
LesnyRumcajs: @samuelarogbonlo they don't have the same issues. You can expect that a version x.y.z of a particular dependency is *always* available from https://rubygems.org/ (unless it gets explicitly yanked). You cannot make such an assumption about dnf and apt packages. From what I know, at least for Alpine, once a version x.y.z+1 is released, the dependency x.y.z is dropped. I recall the same thing for Debian packages, maybe with a longer grace period.

LesnyRumcajs:
@samuelarogbonlo remember that you don't have to make any changes to the `scripts/Dockerfile-snapshot` - it was deleted recently.

.github/workflows/hadolint-ci.yml
samuelarogbonlo: I understand and I think I didn't explain it well. This is what I mean; basically, I never meant to say it was an issue. I only said that we ignored the failing hadolint ci error because 

- firstly the original dockerfiles didn’t have versions pinned to them, pinning versions is actually a best practice but we ignored it because the original dockerfile didn’t have them and also we dont know which exact versions to set.
- The same issue will happen if we are using `apt` and `dnf` and we don't pin versions to them. The wiki is [here](https://github.com/hadolint/hadolint/wiki/DL3028) for ruby - here, you can see that we have to pin the version to resolve the failing CI issues. Now for `apt', check [here](https://github.com/hadolint/hadolint/wiki/DL3008) and see what I mean as well. 

As I said, this is AFAIK, but I might be missing a piece of information or wrongly understanding things.

samuelarogbonlo:
> @samuelarogbonlo remember that you don't have to make any changes to the `scripts/Dockerfile-snapshot` - it was deleted recently.

Ouuu yes yes

.github/workflows/hadolint-ci.yml
```diff
@@ -0,0 +1,21 @@
...
+    steps:
+    - uses: actions/checkout@v2
+
+    - uses: hadolint/hadolint-action@v3.1.0
+      with:
+        dockerfile: "Dockerfile*"
+        recursive: true
+        # https://github.com/hadolint/hadolint/wiki/DL3041
+        # https://github.com/hadolint/hadolint/wiki/DL3050
+        ignore: DL3041,DL3050
```
lemmih, line 21: Missing a newline here. Does your editor automatically remove these?

.github/workflows/hadolint-ci.yml
```diff
@@ -0,0 +1,21 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+
+    - uses: hadolint/hadolint-action@v3.1.0
+      with:
+        dockerfile: "Dockerfile*"
+        recursive: true
+        # https://github.com/hadolint/hadolint/wiki/DL3041
+        # https://github.com/hadolint/hadolint/wiki/DL3050
```
lemmih, line 20: Could you add a comment describing why we ignore these two lints?

lemmih (changes requested):
Do we really need two sets of lock files?

.github/workflows/hadolint-ci.yml
samuelarogbonlo: Maybe cos it seems to work sometimes and doesn't. I'd have to review that and change it.

samuelarogbonlo:
> Do we really need two sets of lock files?

Well, I am not an authority on this but from my dealings; the `gem` specs for both `.lock` files are a little different for the different dockerfiles and function as such. My thoughts are if we use one `.lock` file, then we will have some packages that are needed in one and not in the other - hence my drive to separate the concerns.

LesnyRumcajs:
> > Do we really need two sets of lock files?
> 
> Well, I am not an authority on this but from my dealings; the `gem` specs for both `.lock` files are a little different for the different dockerfiles and function as such. My thoughts are if we use one `.lock` file, then we will have some packages that are needed in one and not in the other - hence my drive to separate the concerns.

I agree with @samuelarogbonlo, those belong to two separate applications. That being said, the structure is a bit confusing, something to be refactored in the future.

lemmih (approved):

.github/workflows/docker.yml
```diff
@@ -38,5 +38,6 @@ jobs:
           file: scripts/Dockerfile-sync-check
           tags: |
             ghcr.io/chainsafe/sync-check:latest
+          labels: ${{ steps.meta.outputs.labels }}
```
LesnyRumcajs, line 41: @samuelarogbonlo To use outputs from the `meta` step, you need a `meta` step.

LesnyRumcajs (dismissed):
Labels won't work in this setup.

.github/workflows/docker.yml
samuelarogbonlo: Just added the step and also change the tags for the build sync step

samuelarogbonlo:
> Labels won't work in this setup.

Just made some changes - what do you think or propose?

PR #58, opened 2023-03-27
chore: restructure forest infrastructure

samuelarogbonlo:
Summary Of Changes 

- Refactor the terraform and Ansible config to accommodate Calibnet and mainnet. 
- Ensure code is Reusable


lemmih:
I'm not a big fan of capitalizing folder names. All our other folders are not capitalized.

Ansible/roles/forest_calib/tasks/main.yml
```diff
@@ -43,22 +43,37 @@
     line: '{{ user }} ALL=(ALL) NOPASSWD: ALL'
     validate: 'visudo -cf %s'
 
+- name: Create docker user
+  ansible.builtin.user:
+    name: '{{ user }}'
+    append: true
+    shell: /bin/bash
+    system: true
+    create_home: false
```
lemmih, line 52: You previously removed this step. Now you've put it back. Could you explain what it does and why we need it?

Ansible/roles/forest_mainnet/templates/config.toml.j2
```diff
@@ -0,0 +1,2 @@
+[client]
+data_dir = "/home/forest/data"
```
lemmih, line 2: Missing final newline.

Ansible/roles/observability/tasks/main.yml
```diff
@@ -12,7 +12,7 @@
     state: directory
     owner: "{{ user }}"
     group: "{{ user }}"
-    mode: "0700"
+    mode: "0700 "
```
lemmih, line 15: Should this extra space be here?

Terraform/calibnet/main.tf
```diff
@@ -0,0 +1,30 @@
+terraform {
+  required_version = ">= 1.3"
+
+  backend "s3" {
+    bucket                      = "forest-test-state"
```
lemmih, line 5: I would prefer if we put all the terraform state files in a single bucket. Otherwise, we end up with lots of buckets. How about you use the `forest-iac` bucket and rename the key to be specific to this service?

Terraform/calibnet/main.tf
```diff
@@ -0,0 +1,30 @@
...
+  name                  = var.name
+  region                = var.region
+  backups               = var.backups
+  image                 = var.image
+  size                  = var.size
+  observability_name    = var.observability_name
+  protocol              = var.protocol
+  source_addresses      = var.source_addresses
+  attach_volume         = false
+  ssh_key               = var.ssh_key
```
lemmih, line 27: Let's use the SSH fingerprints from the entire ChainSafe organization. You can get them like this:
```
data "digitalocean_ssh_keys" "keys" {
  sort {
    key       = "name"
    direction = "asc"
  }
}
```

Terraform/mainnet/terraform.tfvars
```diff
@@ -0,0 +1,19 @@
+# This is a set of values to be referenced as variables in the main code
+
+size                    = "s-4vcpu-8gb"
+image                   = "docker-20-04"
+region                  = "fra1"
+source_addresses        = ["0.0.0.0/0", "::/0"]
+destination_addresses   = ["0.0.0.0/0", "::/0"]
+backups                 = "false"
```
lemmih, line 8: Not sure it ever makes sense to set `backups` to `true` so we can probably just delete this variable.

lemmih (dismissed):
Would it be possible to automatically deploy this infrastructure? That is, automatically deploy it when a PR is merged.

Terraform/calibnet/main.tf
samuelarogbonlo: Yes we could do this but this would mean that I redeploy the whole infrastructure again to add everyone's fingerprints - is that okay? Cos' I thought I had this conversation previously and we agreed it was okay to not have everyone have access to the droplet

samuelarogbonlo:
> Would it be possible to automatically deploy this infrastructure? That is, automatically deploy it when a PR is merged.

Yes, this looks possible but I mean I will have to try that with GitHub secrets to ensure the configurations has access to the values always. I'd test that out and update

samuelarogbonlo:
> Would it be possible to automatically deploy this infrastructure? That is, automatically deploy it when a PR is merged.

Done

ansible/roles/forest_mainnet/tasks/main.yml
```diff
@@ -0,0 +1,108 @@
...
+    recurse: true
+
+- name: Start the Docker container
+  become: true
+  become_user: "{{ user }}"
+  community.general.docker_container:
+    name: forest
+    image: ghcr.io/chainsafe/forest:latest
+    state: started
+    command: --config {{ forest_config }} --encrypt-keystore false --auto-download-snapshot --loki --loki-endpoint "{{ loki_endpoint }}"
```
lemmih, line 81: I don't think `forest_config` should be a variable. You have it hard-coded in this file at line 85.

terraform/Makefile
```diff
@@ -0,0 +1,41 @@
...
+.DEFAULT_GOAL := help
+
+help:
+	@echo "Usage: make [target]"
+	@echo ""
+	@echo "Targets:"
+	@echo "  init      Initialize Terraform"
+	@echo "  plan      Generate and show an execution plan"
+	@echo "  apply     Apply the changes"
+	@echo "  destroy   Destroy the Terraform-managed infrastructure"
```
lemmih, line 16: I don't see these targets in the makefile.

terraform/modules/forest_node/main.tf
```diff
@@ -0,0 +1,90 @@
...
+resource "digitalocean_droplet" "forest" {
+  image    = var.image
+  name     = var.name
+  region   = var.region
+  size     = var.size
+  ssh_keys = data.digitalocean_ssh_keys.keys.ssh_keys.*.fingerprint
+
+  lifecycle {
+    create_before_destroy = true
+  }
```
lemmih, line 45: Is this beneficial?

terraform/modules/forest_node/variable.tf
```diff
@@ -0,0 +1,77 @@
...
+variable "size" {
+  description = "The size of the EC2 instance to launch"
+  type        = string
+}
+
+variable "protocol" {
+  description = "The protocol to use for the connection."
+  type        = string
+}
+
```
lemmih, line 30: When can this be anything other than `tcp`?

terraform/modules/forest_node/variable.tf
```diff
@@ -0,0 +1,77 @@
...
+variable "volume_name" {
+  description = "The name assigned to the volume in the cloud"
+  type        = string
+  default     = ""
+}
+
+variable "project" {
+  description = "The name assigned to the project in the cloud"
+  type        = string
+}
```
lemmih, line 77: Let's default to `Forest-DEV`

terraform/modules/forest_node/variable.tf
```diff
@@ -0,0 +1,77 @@
...
+variable "enviroment" {
+  description = "The environment tag to apply to resources."
+  type        = string
+}
+
+variable "volume_name" {
+  description = "The name assigned to the volume in the cloud"
+  type        = string
+  default     = ""
+}
```
lemmih, line 72: How about we derive this name from the name of the droplet?

lemmih (changes requested):

ansible/roles/forest_mainnet/tasks/main.yml
samuelarogbonlo: Okay noted.

terraform/modules/forest_node/main.tf
samuelarogbonlo: Yes, I believe it is; if we have to shut down some resources and create new ones for any reason, this will minimize service disruptions. So if this resource creation is not done thoroughly, the deletion of old ones cannot happen

terraform/modules/forest_node/variable.tf
samuelarogbonlo: Well, there is UDP and ICMP which one is used for DNS requests settings and stuff and the other is if there is a need for ping requests or something.

terraform/modules/forest_node/variable.tf
samuelarogbonlo: Got it.

.github/workflows/deploy-calibnet.yml
```diff
@@ -0,0 +1,162 @@
...
+        working-directory: ansible
+    env:
+      ANSIBLE_CONFIG: Ansible.cfg
+      AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+      AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+      SLACK_TOKEN: ${{ secrets.SLACK_TOKEN }}
+      SPACES_REGION: "fra1"
+      DOMAIN_NAME: "forest-calibnet.chainsafe.io"
+      SPACES_BUCKET_NAME: "forest-calib-loki-logs"
+      SPACES_ENDPOINT: "fra1.digitaloceanspaces.com"
```
LesnyRumcajs, line 120: This should be configurable, no?

.github/workflows/deploy-calibnet.yml
```diff
@@ -0,0 +1,162 @@
...
+      LETSENCRYPT_EMAIL: "samuel@chainsafe.io"
```
LesnyRumcajs, line 121: Not sure if it's the way to go. Shouldn't we have a group email for this kind of stuff?

.github/workflows/deploy-mainnet.yml
```diff
@@ -0,0 +1,166 @@
...
+
+      - name: Terraform Plan Status
+        if: steps.plan.outcome == 'failure'
+        run: exit 1
+
+      - name: Terraform Apply To Create Forest Infrastructure
+        if: github.ref == 'refs/heads/main' && github.event_name == 'push'
+        run: terraform apply -auto-approve -input=false
+
+      - name: Upload Calibnet Host File As Artifacts
```
LesnyRumcajs, line 100:
```suggestion
      - name: Upload Mainnet Host File As Artifacts
```

.github/workflows/deploy-mainnet.yml
```diff
@@ -0,0 +1,166 @@
...
+        if: github.ref == 'refs/heads/main' && github.event_name == 'push'
+        uses: actions/upload-artifact@v2
+        with:
+          name: hosts_calibnet
+          path: ansible/hosts_mainnet
```
LesnyRumcajs, line 105: so is this calibnet or mainnet?

.github/workflows/deploy-mainnet.yml
```diff
@@ -0,0 +1,166 @@
+---
```
LesnyRumcajs, line 1: the `deploy-calibnet.yml` and `deploy-mainnet.yml` seem almost the same. Seems it's quite prone errors to manually create them. Should we have a single template for those and automatically generate them?

terraform/modules/forest_node/variable.tf
samuelarogbonlo: Well, this works. I'd effect the change

.github/workflows/deploy-mainnet.yml
samuelarogbonlo: Hmm, good thought - this is possible I think. Let me look into it

.github/workflows/deploy-calibnet.yml
```diff
@@ -0,0 +1,57 @@
...
+  configure-forest-infrastructure-calibnet-using-ansible:
+    name: Configure Forest calibnet Infrastructure with Ansible
+    runs-on: ubuntu-latest
+    needs: deploy-forest-calibnet-infrastruture-with-terraform
+    steps:
+      - name: Checkout the code
+        uses: actions/checkout@v3
+
+      # Using Custom Composite action in ./composite-action/ansible folder
+      - name:
```
lemmih, line 45: You are missing a name here.

.github/workflows/deploy-mainnet.yml
```diff
@@ -0,0 +1,57 @@
...
+  configure-forest-infrastructure-mainnet-using-ansible:
+    name: Configure Forest Mainnet Infrastructure with Ansible
+    runs-on: ubuntu-latest
+    needs: deploy-forest-mainnet-infrastruture-with-terraform
+    steps:
+      - name: Checkout the code
+        uses: actions/checkout@v3
+
+                # Using Custom Composite action in ./composite-action/ansible folder
+      - name:
```
lemmih, line 45: You're missing a name here.

terraform/Makefile
lemmih: This help text is wrong. Running `make init` doesn't work.

terraform/calibnet/terraform.tfvars
```diff
@@ -1,15 +1,14 @@
 # This is a set of values to be referenced as variables in the main code
-
 size                  = "s-4vcpu-8gb"
 image                 = "docker-20-04"
 region                = "fra1"
 source_addresses      = ["0.0.0.0/0", "::/0"]
 destination_addresses = ["0.0.0.0/0", "::/0"]
-backups               = "false"
 protocol              = "tcp"
```
lemmih, line 7: I don't understand how `protocol` can be anything other than `tcp`.

terraform/mainnet/terraform.tfvars
```diff
@@ -0,0 +1,17 @@
...
+size                    = "s-4vcpu-8gb"
+image                   = "docker-20-04"
+region                  = "fra1"
+source_addresses        = ["0.0.0.0/0", "::/0"]
+destination_addresses   = ["0.0.0.0/0", "::/0"]
+protocol                = "tcp"
+initial_filesystem_type = "ext4"
+enviroment              = "mainnet"
+project                 = "Forest-DEV"
+volume_size             = "2000"
```
lemmih, line 12: That's a sizeable volume. Let's start with 500.

lemmih (changes requested):
Could you explain how you're updating deployment using ansible? Let's say I change something in the ansible files, how will these changes be deployed to an already-running cluster?

terraform/calibnet/terraform.tfvars
lemmih: You already have rules for UDP and ICMP. When would we want to open port 22 and 1234 for other protocols than `tcp`?

terraform/modules/forest_node/user-data.tpl
```diff
@@ -0,0 +1,90 @@
+#!/bin/bash
+
```
aatifsyed, line 2:
```suggestion
set -euxo pipefail
```
for all bash scripts please
`-e` this exits the script on command failure (rather than you having to check the return code)
'-x` traces commands as they are run (rather than you echoing what you're doing)
'-u` exits on unbound variables
-o pipefail` keeps '-e` for pipes

terraform/modules/forest_node/user-data.tpl
```diff
@@ -0,0 +1,90 @@
...
+NEW_USER="${NEW_USER}"
+USER_DIR="${USER_DIR}"
+USER_DIR_AUTH="${USER_DIR_AUTH}"
+LOKI_ENDPOINT="${LOKI_ENDPOINT}"
+VOLUME_NAME="${VOLUME_NAME}"
+
+# Check if user exists
```
aatifsyed, line 9: remove this check

terraform/modules/forest_node/user-data.tpl
```diff
@@ -0,0 +1,90 @@
...
+if id -u "${NEW_USER}" >/dev/null 2>&1; then
+  echo "User ${NEW_USER} already exists. Exiting..."
+  exit 1
+fi
+
+# Create a login user
+useradd -m ${NEW_USER}
+if [ $? -ne 0 ]; then
```
aatifsyed, line 17: redundant after `-e` applied

terraform/modules/forest_node/user-data.tpl
```diff
@@ -0,0 +1,90 @@
...
+  echo "Failed to create user ${NEW_USER}. Exiting..."
+  exit 1
+fi
+
+# Create .ssh directory for created user
+mkdir -p ${USER_DIR}
```
aatifsyed, line 23:
```suggestion
mkdir --parents -- ${USER_DIR}
```
use longopts for all commands in scripts - it makes it more readable

terraform/modules/forest_node/user-data.tpl
```diff
@@ -0,0 +1,90 @@
...
+chown ${NEW_USER}:${NEW_USER} ${USER_DIR}
```
aatifsyed, line 24: you should quote variables - can you run this through `shellcheck` please, it should pick up on stuff like this. Alternatively, use the shellcheck extension for vscode

terraform/modules/forest_node/user-data.tpl
```diff
@@ -0,0 +1,90 @@
+#!/bin/bash
+
+NEW_USER="${NEW_USER}"
```
aatifsyed, line 3: you don't need to quote these vars. Also, why are you doing this assignment?

terraform/modules/forest_node/user-data.tpl
```diff
@@ -0,0 +1,90 @@
...
+
+# Add new user to "docker" group
+usermod -aG docker ${NEW_USER}
+
+# check if the volume name is defined
+if [ -n "${VOLUME_NAME}" ]; then
+
+  # set-up forest volume 
+  mkdir -p /mnt/forest_mainnet_volume
+  mount -o discard,defaults,noatime /dev/disk/by-id/scsi-0DO_Volume_forest-mainnet-volume /mnt/forest_mainnet_volume
```
aatifsyed, line 58: what do these options mean? Why are you adding them?

terraform/modules/forest_node/user-data.tpl
```diff
@@ -0,0 +1,90 @@
...
+  cp /root/.ssh/authorized_keys ${USER_DIR_AUTH}
+  chown ${NEW_USER}:${NEW_USER} ${USER_DIR_AUTH}
+  chmod 0600 ${USER_DIR_AUTH}
+fi
+
+# Restart SSHD
+systemctl restart sshd
+
+# Allow created user to log in
+echo "AllowUsers ${NEW_USER}" >> /etc/ssh/sshd_config
```
aatifsyed, line 41: shouldn't this be before restarting sshd?

terraform/modules/forest_node/user-data.tpl
```diff
@@ -0,0 +1,90 @@
...
+if [ $? -ne 0 ]; then
+  echo "Failed to update sudoers file. Exiting..."
+  exit 1
+fi
+
+# Add new user to "docker" group
+usermod -aG docker ${NEW_USER}
+
+# check if the volume name is defined
+if [ -n "${VOLUME_NAME}" ]; then
```
aatifsyed, line 54: don't need to do this with `-u`

terraform/modules/forest_node/user-data.tpl
```diff
@@ -0,0 +1,90 @@
...
+
+  # set-up forest volume 
+  mkdir -p /mnt/forest_mainnet_volume
+  mount -o discard,defaults,noatime /dev/disk/by-id/scsi-0DO_Volume_forest-mainnet-volume /mnt/forest_mainnet_volume
+
+  # Change ownership of volume directory
+  chown -R 1000:1000 /mnt/${VOLUME_NAME}
```
aatifsyed, line 61: why 1000? comment please

aatifsyed (changes requested):
I've had a look at the bash script. I'd like to re-review once you've made the changes

aatifsyed (changes requested):
I've had a look at the bash script. I'd like to re-review once you've made the changes

terraform/modules/forest_node/user-data.tpl
```diff
@@ -0,0 +1,90 @@
...
+  mkdir -p /etc/${NEW_USER} && chmod 0777 /etc/${NEW_USER}
+
+  # Create the config.toml file
+  cat << EOF > /etc/${NEW_USER}/config.toml
+[client]
+data_dir = "/home/${NEW_USER}/data"
+EOF
+
+  # Start the Forest  Mainnet Docker container with volume
+  su - ${NEW_USER} -c "docker run -d --name forest -v /etc/${NEW_USER}/config.toml:/home/${NEW_USER}/config.toml -v /mnt/$VOLUME_NAME:/home/${NEW_USER}/data -p 1234:1234 -p 6116:6116 -p 3100:3100 --restart always ghcr.io/chainsafe/forest:v0.8.2 --config /home/${NEW_USER}/config.toml --encrypt-keystore false --auto-download-snapshot --loki --loki-endpoint \"${LOKI_ENDPOINT}\""
```
lemmih, line 73: Why pin to version `0.8.2`?

terraform/modules/forest_node/main.tf
```diff
@@ -0,0 +1,94 @@
...
+  image    = var.image
+  name     = var.name
+  region   = var.region
+  size     = var.size
+  ssh_keys = data.digitalocean_ssh_keys.keys.ssh_keys.*.fingerprint
+
+  user_data = templatefile("${path.module}/user-data.tpl", {
+    NEW_USER      = var.name
+    USER_DIR      = "/home/${var.name}/.ssh"
+    USER_DIR_AUTH = "/home/${var.name}/.ssh/authorized_keys"
```
lemmih, line 46: I don't understand the purpose of these two parameters. When would we ever want to change them?

.github/workflows/deploy-calibnet.yml
```diff
@@ -0,0 +1,57 @@
...
+
+      # Using Custom Composite action in ./composite-action/ansible folder
+      - name: Calibnet Ansible Composite
+        uses: ./composite-action/ansible
+        with:
+          working_directory: ansible
+          environment: calibnet
+          domain_name: forest-calibnet.chainsafe.io
+          spaces_bucket_name: forest-calib-loki-logs
+          letsencrypt_email: samuel@chainsafe.io
```
lemmih, line 52: Why is `letsencrypt_email` set for calibnet but not for mainnet?

ansible/letsencrypt.yaml
```diff
@@ -6,5 +6,5 @@
   roles:
     - letsencrypt
   vars:
-    letsencrypt_email: ""
+    letsencrypt_email: "samuel@chainsafe.io"
```
lemmih, line 9: Can we use the value set in the workflow?

terraform/modules/forest_node/user-data.tpl
aatifsyed: as in, you're assigning a variable to itself, which is a no-op, you can remove all lines like this

terraform/modules/forest_node/user-data.tpl
```diff
@@ -0,0 +1,65 @@
+#!/bin/bash
```
aatifsyed, line 1: Why is this a `.tpl` file - who's preprocessing it? Please comment

terraform/modules/forest_node/user-data.tpl
```diff
@@ -0,0 +1,65 @@
...
+
+set -euxo pipefail
+
+NEW_USER="${NEW_USER}"
+VOLUME_NAME="${VOLUME_NAME}"
+
+# Create a login user
+useradd -m "${NEW_USER}"
```
aatifsyed, line 9: please use longopts where possible

terraform/modules/forest_node/user-data.tpl
```diff
@@ -0,0 +1,65 @@
...
+
+# Check if authorized_keys file exists
+if [ ! -f "/root/.ssh/authorized_keys" ]; then
+  echo "No authorized_keys file found. Skipping copying..."
+else
+  # Copy public key from existing root to new user
+  cp /root/.ssh/authorized_keys "/home/${NEW_USER}/.ssh/authorized_keys"
+  chown "${NEW_USER}:${NEW_USER}" "/home/${NEW_USER}/.ssh/authorized_keys"
+  chmod 0600 "/home/${NEW_USER}/.ssh/authorized_keys"
+fi
```
aatifsyed, line 24:
```suggestion
if [ -f "/root/.ssh/authorized_keys" ]; 
  : Copy public key from existing root to new user
  cp /root/.ssh/authorized_keys "/home/${NEW_USER}/.ssh/authorized_keys"
  chown "${NEW_USER}:${NEW_USER}" "/home/${NEW_USER}/.ssh/authorized_keys"
  chmod 0600 "/home/${NEW_USER}/.ssh/authorized_keys"
fi
```

terraform/modules/forest_node/user-data.tpl
```diff
@@ -0,0 +1,65 @@
...
+systemctl restart sshd
+
+# Add user to sudoers file after checking visudo
+if ! echo "${NEW_USER} ALL=(ALL) NOPASSWD: ALL" | (EDITOR="tee -a" visudo); then
+  echo "Failed to update sudoers file. Exiting..."
+  exit 1
+fi
+
+# Add new user to "docker" group
+usermod -aG docker "${NEW_USER}"
```
aatifsyed, line 39:
```suggestion
usermod --append --groups docker "${NEW_USER}"
```

terraform/modules/forest_node/user-data.tpl
```diff
@@ -0,0 +1,65 @@
...
+echo "AllowUsers ${NEW_USER}" >> /etc/ssh/sshd_config
+
+# Restart SSHD
+systemctl restart sshd
+
+# Add user to sudoers file after checking visudo
+if ! echo "${NEW_USER} ALL=(ALL) NOPASSWD: ALL" | (EDITOR="tee -a" visudo); then
+  echo "Failed to update sudoers file. Exiting..."
+  exit 1
+fi
```
aatifsyed, line 36:
```suggestion
: Add user to sudoers file # We pipe to visudo to check syntax
echo "${NEW_USER} ALL=(ALL) NOPASSWD: ALL" | (EDITOR="tee --append" visudo)
```

terraform/modules/forest_node/user-data.tpl
```diff
@@ -0,0 +1,65 @@
...
+# check if the volume name is defined
+# If the volume name is defined, start the Forest Mainnet Docker container with volume
+if [ -n "${VOLUME_NAME}" ]; then
+
+  # set-up forest volume for mainnet container
+  mkdir --parents -- /mnt/"${VOLUME_NAME}"
+  mount -o discard,defaults,noatime /dev/disk/by-id/scsi-0DO_Volume_forest-mainnet-volume /mnt/"${VOLUME_NAME}"
+
+  # Change ownership of volume directory to the created user
+  chown -R "${NEW_USER}":"${NEW_USER}" /mnt/"${VOLUME_NAME}"
```
aatifsyed, line 50:
```suggestion
  chown --recursive "${NEW_USER}":"${NEW_USER}" /mnt/"${VOLUME_NAME}"
```

terraform/modules/forest_node/user-data.tpl
```diff
@@ -0,0 +1,65 @@
...
+
+  # Make sure /etc/forest config directory exists
+  mkdir --parents -- /etc/"${NEW_USER}" && chmod 0777 /etc/"${NEW_USER}"
+
+  # Create the config.toml file
+  cat << EOF > /etc/"${NEW_USER}"/config.toml
+[client]
+data_dir = "/home/${NEW_USER}/data"
```
aatifsyed, line 58: will the quotes be in the toml file or not? i.e is bash doing string interpolation

terraform/modules/forest_node/user-data.tpl
```diff
@@ -0,0 +1,65 @@
...
+
+  # set-up forest volume for mainnet container
+  mkdir --parents -- /mnt/"${VOLUME_NAME}"
+  mount -o discard,defaults,noatime /dev/disk/by-id/scsi-0DO_Volume_forest-mainnet-volume /mnt/"${VOLUME_NAME}"
+
+  # Change ownership of volume directory to the created user
+  chown -R "${NEW_USER}":"${NEW_USER}" /mnt/"${VOLUME_NAME}"
+
+  # Make sure /etc/forest config directory exists
+  mkdir --parents -- /etc/"${NEW_USER}" && chmod 0777 /etc/"${NEW_USER}"
```
aatifsyed, line 53: the line doesn't match the comment

terraform/modules/forest_node/user-data.tpl
```diff
@@ -0,0 +1,65 @@
...
+  exit 1
+fi
+
+# Add new user to "docker" group
+usermod -aG docker "${NEW_USER}"
+
+# check if the volume name is defined
+# If the volume name is defined, start the Forest Mainnet Docker container with volume
+if [ -n "${VOLUME_NAME}" ]; then
+
```
aatifsyed, line 44:
```suggestion
if [ -n "${VOLUME_NAME}" ]; then
    : mounting user-specified volume for use with forest container
```

terraform/modules/forest_node/user-data.tpl
```diff
@@ -0,0 +1,65 @@
...
+  # set-up forest volume for mainnet container
+  mkdir --parents -- /mnt/"${VOLUME_NAME}"
+  mount -o discard,defaults,noatime /dev/disk/by-id/scsi-0DO_Volume_forest-mainnet-volume /mnt/"${VOLUME_NAME}"
```
aatifsyed, line 47:
```suggestion
  # discard: notify the volume to free blocks (useful for SSDs)
  # defaults: default mount options, including rw
  # noatime: don't preserve file access times
  mount --options discard,defaults,noatime /dev/disk/by-id/scsi-0DO_Volume_forest-mainnet-volume /mnt/"${VOLUME_NAME}"
```

terraform/modules/forest_node/user-data.tpl
```diff
@@ -0,0 +1,65 @@
...
+  # Create the config.toml file
+  cat << EOF > /etc/"${NEW_USER}"/config.toml
+[client]
+data_dir = "/home/${NEW_USER}/data"
+EOF
+
+  su - "${NEW_USER}" -c "docker run -d --name forest -v /etc/${NEW_USER}/config.toml:/home/${NEW_USER}/config.toml -v /mnt/$VOLUME_NAME:/home/${NEW_USER}/data -p 1234:1234 --restart always ghcr.io/chainsafe/forest:latest --config /home/${NEW_USER}/config.toml --encrypt-keystore false --auto-download-snapshot"
+else
+  # If the volume name is not defined, start the Forest Calibnet Docker container
+  su - "${NEW_USER}" -c "docker run -d --name forest -p 1234:1234 --restart always ghcr.io/chainsafe/forest:latest --encrypt-keystore false --auto-download-snapshot --chain calibnet "
```
aatifsyed, line 64: these commands are a bit hard to read - can you restructure this as e.g a var to highlight the actual differences?

aatifsyed (changes requested):

terraform/modules/forest_node/user-data.tpl
segunjkf: this is a good thought but i think you're missing `then` i.e `if [ -f "/root/.ssh/authorized_keys" ]; then`

terraform/modules/forest_node/user-data.sh
```diff
@@ -0,0 +1,81 @@
...
+    -p 1234:1234 \
+    --restart always \
+    ghcr.io/chainsafe/forest:latest \
+    --config /home/${NEW_USER}/config.toml \
+    --encrypt-keystore false \
+    --auto-download-snapshot"
+
+else
+  # If the volume name is not defined, start the Forest Calibnet Docker container
+  su - "${NEW_USER}" -c "docker run -d --name forest \
```
aatifsyed, line 74: recommend `doas` instead of constructing a command string like this

terraform/modules/forest_node/user-data.sh
```diff
@@ -0,0 +1,81 @@
+#!/bin/bash
+
+set -euxo pipefail
+
+NEW_USER="sam"
+
+VOLUME_NAME="forest_mainnet_volume"
```
aatifsyed, line 7: these don't look right

terraform/modules/forest_node/user-data.sh
```diff
@@ -0,0 +1,81 @@
...
+
+# Create a login user
+useradd -m "${NEW_USER}"
```
aatifsyed, line 10:
```suggestion
useradd --create-home -- "$NEW_USER"
```

Also, why have you gone for `${FOO}` syntax instead of `$FOO`? It's unusual

terraform/modules/forest_node/user-data.sh
```diff
@@ -0,0 +1,81 @@
...
+  # set-up forest volume for mainnet container
+  mkdir --parents -- /mnt/"${VOLUME_NAME}"
+
+  # discard: notify the volume to free blocks (useful for SSDs)
+  # defaults: default mount options, including rw
+  # noatime: don't preserve file access times
+  mount --options discard,defaults,noatime /dev/disk/by-id/scsi-0DO_Volume_forest-mainnet-volume /mnt/"${VOLUME_NAME}"
+
+  # Change ownership of volume directory to the created user
+  chown -R "${NEW_USER}":"${NEW_USER}" /mnt/"${VOLUME_NAME}"
```
aatifsyed, line 51:
```suggestion
  chown --recursive "${NEW_USER}":"${NEW_USER}" /mnt/"${VOLUME_NAME}"
```

terraform/modules/forest_node/user-data.sh
```diff
@@ -0,0 +1,81 @@
...
+
+  su - "${NEW_USER}" -c "docker run -d --name forest \
+    -v /etc/${NEW_USER}/config.toml:/home/${NEW_USER}/config.toml \
+    -v /mnt/$VOLUME_NAME:/home/${NEW_USER}/data \
+    -p 1234:1234 \
+    --restart always \
+    ghcr.io/chainsafe/forest:latest \
+    --config /home/${NEW_USER}/config.toml \
+    --encrypt-keystore false \
+    --auto-download-snapshot"
```
aatifsyed, line 70: better layout, but you can still dedup between them
```bash
command=(
  docker
  run
  --detach
  --name=forest
  --port=1234:1234
  --restart=always
  # etc
)

if [ -n $VOLUME_NAME ]; then
    command+=(
         --volume=...
    )
fi

"${command[@]}"
```

aatifsyed (dismissed):
These are more nits at this point, apart from the top suggestion - feel fee to dismiss if you're on a tight budget, but otherwise I think we can keep higher quality scripts :)

terraform/Makefile
```diff
@@ -0,0 +1,41 @@
...
+
+help:
+	@echo "Usage: make [target]"
+	@echo ""
+	@echo "Targets:"
+	@echo "  init_calib or init_main      	Initialize Terraform for either mainnet or calibnet"
+	@echo "  plan_calib or plan_main     	Generate and show an execution plan for either mainnet or calibnet"
+	@echo "  apply_main or apply_calib  	Apply the changes for either mainnet or calibnet"
+	@echo "  destroy_main or destroy_calib  Destroy the Terraform-managed infrastructure for either mainnet or calibnet"
+	@echo "  help      						Shows this help message"
```
lemmih, line 17: You're mixing tabs and spaces. Always use spaces. Otherwise, the indentation will not be the same for everyone.

terraform/modules/forest_node/user-data.sh
```diff
@@ -0,0 +1,75 @@
...
+# If the volume name is defined, start the Forest Mainnet Docker container with volume
+if [ -n "${VOLUME_NAME}" ]; then
+
+  # set-up forest volume for mainnet container
+  mkdir --parents -- /mnt/"${VOLUME_NAME}"
+
+  # discard: notify the volume to free blocks (useful for SSDs)
+  # defaults: default mount options, including rw
+  # noatime: don't preserve file access times
+  mount --options discard,defaults,noatime /dev/disk/by-id/scsi-0DO_Volume_forest-mainnet-volume /mnt/"${VOLUME_NAME}"
```
lemmih, line 42: You've hard-coded `forest-mainnet-volume` here.

terraform/modules/forest_node/user-data.sh
```diff
@@ -0,0 +1,75 @@
...
+
+else
+  # If the volume name is not defined, start the Forest Calibnet Docker container
+  su - "${NEW_USER}" -c "docker run -d --name forest \
+    -p 1234:1234 \
+    --restart always \
+    ghcr.io/chainsafe/forest:latest \
+    --encrypt-keystore false \
+    --auto-download-snapshot \
+    --chain calibnet"
```
lemmih, line 74: The chain should be an explicit parameter in the terraform module rather than inferred by the volume's presence or absence.

lemmih (changes requested):
We are getting close now. We also need something like `watchtower` to keep the docker images up-to-date, but I'm okay with putting that in another PR.

lemmih (changes requested):

scripts/sync_check/boot_service.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# The .env file contains environment variables that we want access to.
+# The .env file contains the chain variables that we want access to.
```
lemmih, line 3: I think you accidentally did a search&replace across the entire repository. You shouldn't be modifying this file.

lemmih (changes requested):

lemmih (commented):
The README is outdated, and the docker container isn't automatically kept up-to-date. But I'm okay with those things being handled in a follow-up PR. Let's have @aatifsyed go over the bash script once more before merging.

terraform/modules/forest_node/user-data.sh
```diff
@@ -0,0 +1,64 @@
+#!/bin/bash
+
```
aatifsyed, line 2:
```suggestion
# This bash script is templated using <engine>, with the variables defined in <tfvars>.
# THIS MEANS THAT ${VARIABLES} YOU SEE ARE FOR THE TEMPLATE ENGINE, NOT BASH
#
# It is inserted to the newly created VM at <some path> as `user-data` by <orchestration>, and run <at startup>.
# TODO(samuelarogbonlo): refactor so we don't template the script itself.
```

I'm guessing at a lot of stuff here from the filename, and previous discussion, but this is the kind of information that should be contained in the code (or the README if it's really basic).

You've probably had to learn a lot about `terraform` and orchestration to write this code - why not save that knowledge for future you, or the next developer or reviewer ;)
That way, our code stays easier to maintain, review etc.

I'm also still not convinced by templating a bash script like this - scripts take _arguments_, they're not _templated_

terraform/modules/forest_node/user-data.sh
```diff
@@ -0,0 +1,64 @@
...
+set -euxo pipefail
+
+# Create a login user
```
aatifsyed, line 5:
```suggestion
: create a user
# --disabled-password: the only way a user can log in is with e.g ssh
# --gecos "": don't set address and phone number
```

Print out the description as the command runs - very useful for debugging remotely failing scripts ;)

Use a comment to save information for future you/other readers.

If a reasonable reader needs to ask "why have they done it like this", it's a good place for a comment


Why have you manually set `--home` to the default value - please comment the script or remove

terraform/modules/forest_node/user-data.sh
```diff
@@ -0,0 +1,64 @@
...
+adduser --home "/home/${NEW_USER}" --disabled-password --gecos "" "${NEW_USER}"
+
+# Create .ssh directory for created user
```
aatifsyed, line 8:
```suggestion
: Create .ssh directory with the permissions required by `ssh`
```

terraform/modules/forest_node/user-data.sh
```diff
@@ -0,0 +1,64 @@
...
+mkdir --parents -- "/home/${NEW_USER}/.ssh"
+chown "${NEW_USER}:${NEW_USER}" "/home/${NEW_USER}/.ssh"
+chmod 0700 "/home/${NEW_USER}/.ssh"
+
+# Check if authorized_keys file exists and copy it from root to new user
+if [ -f "/root/.ssh/authorized_keys" ]; then
+  cp /root/.ssh/authorized_keys "/home/${NEW_USER}/.ssh/authorized_keys"
+  chown "${NEW_USER}:${NEW_USER}" "/home/${NEW_USER}/.ssh/authorized_keys"
```
aatifsyed, line 16:
```suggestion
if [ -f "/root/.ssh/authorized_keys" ]; then
  : Allowing those with root ssh keys to log in as ${NEW_USER}
  cp /root/.ssh/authorized_keys "/home/${NEW_USER}/.ssh/authorized_keys"
  chown "${NEW_USER}:${NEW_USER}" "/home/${NEW_USER}/.ssh/authorized_keys"
```
Why are we doing things this way - could you comment the file?
This is unusual to me, because if someone already has root access, then they could trivially access the `user` login. One explanation is that we always want to run forest as `user` instead of `root`.

terraform/modules/forest_node/user-data.sh
```diff
@@ -0,0 +1,64 @@
...
+  chmod 0600 "/home/${NEW_USER}/.ssh/authorized_keys"
+fi
+
+# Allow created user to log in
+echo "AllowUsers ${NEW_USER}" >> /etc/ssh/sshd_config
+
+# Restart SSHD
+systemctl restart sshd
+
```
aatifsyed, line 25:
```suggestion
: Allow user to log in via ssh
echo "AllowUsers ${NEW_USER}" >> /etc/ssh/sshd_config
systemctl restart sshd

```

A lot of your comments so far are of the form
```bash
# do the thing
do-the-thing
```

Which isn't very helpful - comments are to help the reader:
- understand the layout of the script (large sets of related actions which)
- answer reader questions ahead of time (less obvious knowledge)

the line `systemctl restart sshd` _already tells me_ that you're restarting sshd, you don't need to do that twice

terraform/modules/forest_node/user-data.sh
```diff
@@ -0,0 +1,64 @@
...
+# Add user to sudoers file. We pipe to visudo to check syntax
+usermod --append --groups sudo "${NEW_USER}"
+
+# Add new user to "docker" group
+usermod --append --groups docker "${NEW_USER}"
+
```
aatifsyed, line 31:
```suggestion
# We want the user to be able to run a docker container, and to have general admin rights
usermod --append --groups sudo,docker "${NEW_USER}"

```

aatifsyed: I'm not really able to audit the security of these changes because I don't have an idea of _why_ this file exists, what the VM is for, etc. The README doesn't really tell me - where can future readers find this information?

terraform/modules/forest_node/user-data.sh
```diff
@@ -0,0 +1,64 @@
...
+# set-up forest data directory for Forest container
+mkdir --parents -- "/home/${NEW_USER}/forest_data"
+
+# Check if the volume name is defined
+# If the volume name is defined, mount the volume
+if [ -n "${VOLUME_NAME}" ]; then
```
aatifsyed, line 37:
```suggestion
if [ -n "${VOLUME_NAME}" ]; then
```

terraform/modules/forest_node/user-data.sh
```diff
@@ -0,0 +1,64 @@
...
+data_dir = "/home/${NEW_USER}/forest_data/data"
+EOF
+
+# Run the Forest Docker container
+su - "${NEW_USER}" -c "docker run -d --name forest \
+  -v /home/${NEW_USER}/forest_data:/home/${NEW_USER}/data \
+  -p 1234:1234 \
+  --restart always \
+  ghcr.io/chainsafe/forest:latest \
+  --config /home/${NEW_USER}/data/config.toml \
```
aatifsyed, line 61: the `--config` path doesn't match the file you created in line 50

terraform/modules/forest_node/user-data.sh
```diff
@@ -0,0 +1,64 @@
...
+  --encrypt-keystore false \
+  --auto-download-snapshot \
+  --chain ${CHAIN}"
```
aatifsyed, line 64:
```suggestion
sudo --user="${NEW_USER}" -- \
  docker \ 
  run \
  --detach \
  --name=forest \
  --volume=/home/${NEW_USER}/forest_data:/home/${NEW_USER}/data \
  --publish=1234:1234 \
  --restart=always \
  ghcr.io/chainsafe/forest:latest \
  --config=/home/${NEW_USER}/data/config.toml \
  --encrypt-keystore=false \
  --auto-download-snapshot \
  --chain=${CHAIN}
```

I think I mentioned this in a previous review - rather than sticking your arguments in a string, you can keep them flat using `sudo` or `doas` - you should only stoop to constructing a string if you _need_ shell functionality, but you're in a shell script so you don't.

This is also the correct place to create an array of args in slack if you need, but looks like you've refactored that stuff away - nice :)
It would also be less ugly to keep them in an array so you don't need to escape all the newlines, so feel free to make that change

terraform/modules/forest_node/user-data.sh
```diff
@@ -0,0 +1,64 @@
...
+
+  # mount volume at the forest data directory
+  # discard: notify the volume to free blocks (useful for SSDs)
+  # defaults: default mount options, including rw
+  # noatime: don't preserve file access times
+  mount --options discard,defaults,noatime /dev/disk/by-id/scsi-0DO_Volume_"${DISK_ID_VOLUME_NAME}" "/home/${NEW_USER}/forest_data"
+fi
+
+# Change ownership of directory to the created user
+chown -R "${NEW_USER}":"${NEW_USER}" "/home/${NEW_USER}/forest_data"
```
aatifsyed, line 47:
```suggestion
chown --recursive "${NEW_USER}":"${NEW_USER}" "/home/${NEW_USER}/forest_data"
```

aatifsyed (approved):
There are a couple of nits/possible bugs, but the code looks fine.
The substance of my comments are about maintainability and documentation, but getter better at those happens over a few reviews, not all at once.
It might be better to leave some docs for future Samuel

PR #60, opened 2023-03-28
fix: enable ssh-agent only when deploying

lemmih:
**Summary of changes**
- The SSH key isn't available for first-time contributors. Change the workflow such that the ssh key is only required when we actually deploy (merge to main).

Closes 


lemmih:
Hmm, it turns out that GitHub is having issues with secrets. I still think this is a good PR but it might not fix the underlying issue.

PR #79, opened 2023-05-08
feat: upload compressed snapshots

lemmih:
**Summary of changes**
- Include compressed files in purges
- Generate and upload compressed snapshots (in addition to regular snapshots)
- Opened an issue to fix the paths in checksum files: https://github.com/ChainSafe/forest/issues/2835

Closes 


daily_snapshot_terraform/modules/daily_snapshot/service/list_snapshots.rb
```diff
@@ -24,7 +26,7 @@ def list_snapshots(chain_name = 'calibnet', bucket = 'forest-snapshots', endpoin
       file.match(snapshot_format) { |m| 
         url = "https://#{bucket}.#{endpoint}/#{chain_name}/#{file}"
         url_s3 = "s3://#{bucket}/#{chain_name}/#{file}"
-        snapshot = Snapshot.new m[:network], m[:date].to_date, m[:height].to_i, m[:compressed], file, url_s3, url
+        snapshot = Snapshot.new m[:network], m[:date].to_date, m[:height].to_i, file, url_s3, url
```
LesnyRumcajs, line 29: It might be worth making `Snapshot` a proper class to avoid such unexpected (`date.to_date`) conversions. It will be easier to extend it in the future as well. What do you think?

lemmih: I'm at the very edge of my Ruby skills. :) I struggle to grasp the differences between a struct and a class. Mind helping me convert the code to use a class?

LesnyRumcajs: I find a Ruby `Struct` a shorthand for a simplified class, just when you want to combine some things. Kind of like a named tuple. Let me try to re-write it a bit.

LesnyRumcajs: @lemmih I've pushed a proposition of how it might look like. Let me know what you think and please check if it works. It should but no promises.

lemmih: Nice, it works. Would it be okay to define `to_s` for `Snapshot`? Setting `to_s` to be the URL helped me check that it worked.

LesnyRumcajs: Sure. Would you like me to add it or do you already have the code?

lemmih: I'll add it.

lemmih: Pushed.

LesnyRumcajs (approved):

PR #80, opened 2023-05-15
feat: terrform the sync check service

### LINKED ISSUE #81: Terraform the sync-check service
**Issue summary**

The sync-chain service uses the Docker-in-Docker approach and cannot provision its hardware. We should use Terraform to provision and deploy the service on DigitalOcean automatically.

lemmih:
**Summary of changes**
- 

Closes #81 


LesnyRumcajs (changes requested):
Please add a `README` along the lines of https://github.com/ChainSafe/forest-iac/blob/main/daily_snapshot_terraform/README.md

lemmih (commented):
- The service isn't automatically updated when the source files are modified. You need to tell terraform that the service depends on the source files.
- Keep the `remote-exec` commands as small as possible. Any changes to these commands do not trigger a re-deployment so it's important to move as much as possible over to an init script.
- The service isn't properly restored if the droplet is rebooted, I think. If I'm wrong about this, add documentation that explains how it works.
- You still need watchtower to keep the Forest image up-to-date.
- There should be a Github workflow to automatically deploy this service when it is merged into `main`.

LesnyRumcajs:
@sudo-shashank could you please remove the original files (from `scripts/`) that you moved?

sudo-shashank:
@LesnyRumcajs do we still need `Build Docker Infrastructure image and push to repositories` in CI workflow ?

sync_check_terraform/.gitignore
LesnyRumcajs, line 1: How about consolidating all those ignores files in the root `.gitignore`? 

We already have it https://github.com/ChainSafe/forest-iac/blob/main/daily_snapshot_terraform/.gitignore and https://github.com/ChainSafe/forest-iac/blob/main/forest-node/terraform/.gitignore

sync_check_terraform/main.tf
```diff
@@ -0,0 +1,178 @@
...
+    # validation.
+    skip_credentials_validation = "true"
+  }
+}
+
+locals {
+  name                  = "forest-sync-check" # droplet name
+  size                  = "so-2vcpu-16gb"     # droplet size
+  image                 = "fedora-36-x64"
+  slack_channel         = "#forest-notifications" # slack channel for notifications
```
LesnyRumcajs, line 28: Can we keep those variables in the same way as in the daily snapshot? I don't know which one is the "preferred" way of doing things in Terraform, but it'd be nice to have a common service structure in the repo so it is easy to switch between them cognitively. 

https://github.com/ChainSafe/forest-iac/blob/main/daily_snapshot_terraform/main.tf

sync_check_terraform/service/restart.service
```diff
@@ -0,0 +1,9 @@
+[Unit]
```
LesnyRumcajs, line 1: Why do you need to do this? Wouldn't a single `systemctl enable docker` be enough in the droplet initialisation? Or am I missing something?

sync_check_terraform/service/docker-compose.yml
```diff
@@ -94,6 +94,7 @@ services:
     depends_on:
       - forest_mainnet
       - forest_calibnet
+    restart: unless-stopped
```
LesnyRumcajs, line 97: Why?

sync_check_terraform/service/Dockerfile-tester
```diff
@@ -21,3 +21,6 @@ WORKDIR /chainsafe
 COPY Gemfile Gemfile.lock ./
 
 RUN gem install bundler && bundle install
+
+COPY ruby_common ruby_common
+COPY * ./
```
LesnyRumcajs, line 26: Are you sure you want to copy over everything? If so, the first line seems redundant, no?

sync_check_terraform/main.tf
```diff
@@ -0,0 +1,178 @@
...
+resource "digitalocean_firewall" "forest-firewall" {
+  name = local.name
+
+  inbound_rule {
+    protocol         = "tcp"
+    port_range       = "22"
+    source_addresses = local.source_addresses
+  }
+
+  inbound_rule {
```
LesnyRumcajs, line 142: why do we need to have it exposed?

sync_check_terraform/main.tf
```diff
@@ -0,0 +1,178 @@
...
+      "echo 'export FOREST_TARGET_RUBY_COMMON=/volumes/ruby_common' >> .forest_env",
+      "echo 'export FOREST_SLACK_API_TOKEN=\"${var.slack_token}\"' >> .forest_env",
+      "echo 'export FOREST_SLACK_NOTIF_CHANNEL=\"${local.slack_channel}\"' >> .forest_env",
+      "echo 'source .forest_env' >> .bashrc",
+      "source ~/.forest_env",
+      "/bin/bash ./init.sh > init_log.txt",
+      "nohup /bin/bash ./run_service.sh > run_service_log.txt &",
+      "cp ./restart.service /etc/systemd/system/",
+      "systemctl enable restart.service",
+      # Exiting without a sleep sometimes kills the script :-/
```
LesnyRumcajs, line 116: Do we know why? I mean, I'm okay with the workaround, but it'd be nice to know the reason.

LesnyRumcajs (changes requested):

sync_check_terraform/.gitignore
sudo-shashank: done

sync_check_terraform/main.tf
sudo-shashank: done

sync_check_terraform/service/restart.service
sudo-shashank: we do `systemctl enable docker` during droplet initialisation but on reboot we need to start the docker service again.
this is just a service to restart the docker service on reboot

sync_check_terraform/service/docker-compose.yml
sudo-shashank: using this we can restart the sync check after droplet reboots, docker service is restarted on droplet reboot which starts `forest-calibnet`, `forest-mainnet` and `forest-tester` and sync check restarts on droplet reboot

sync_check_terraform/service/restart.service
LesnyRumcajs: I don't understand, why? `systemctl enable docker` should handle reboots just fine.
https://docs.docker.com/engine/install/linux-postinstall/#configure-docker-to-start-on-boot-with-systemd

sync_check_terraform/service/Dockerfile-tester
sudo-shashank: fixed using `COPY . .` which copies all files and dir recursively maintaining the folder structure

sync_check_terraform/service/docker-compose.yml
LesnyRumcajs: What about the logic contained in the `run_service.sh`, including the `sync_check_process.rb`? Will this process, monitoring disk usage, be spawned upon reboot?

sync_check_terraform/main.tf
sudo-shashank: no removed

sync_check_terraform/service/docker-compose.yml
sudo-shashank: as per my understanding, we don't need to run `sync_check_process.rb` on reboot instead `forest-tester` commands get re-run on restart of docker services which include `sync_check.rb forest-calibnet` and `sync_check.rb forest-mainnet`

sync_check_terraform/service/restart.service
sudo-shashank: yes right, removed restart service

sync_check_terraform/service/docker-compose.yml
LesnyRumcajs: Why don't we need to run it after a reboot? Say the machine gets rebooted - we still want to monitor disk usage and do the cleanup, no? This logic is not contained in the `forest-tester`.

While we do have garbage collection, we still need to do a cleanup from time to time (once a month or so, from what I see).

sync_check_terraform/service/Dockerfile-tester
LesnyRumcajs: But do we need everything? Why does the tester need, e.g., `docker-compose.yml`?

sudo-shashank: no we don't, changed to copy only required files

sync_check_terraform/service/docker-compose.yml
LesnyRumcajs: if we re-run `run_service.sh` at reboot it will cleanup the disk with the `sync_check_process.rb`.

I don't quite get why we moved away from the original Docker container - it had restarts in-built. :)

Alternatively we can have a restart service that you introduced before, just not restarting docker but the `run_service.sh`.

Did I answer your question?

sudo-shashank: yes

sudo-shashank: done

.github/workflows/deploy-sync-check.yml
```diff
@@ -0,0 +1,53 @@
...
+
+      - name: Terraform Validate
+        id: validate
+        run: terraform validate -no-color
+      
+      - name: Terraform Plan
+        id: plan
+        if: github.event_name == 'pull_request'
+        run: terraform plan -no-color -input=false
+        continue-on-error: true
```
LesnyRumcajs, line 43: why do we want to continue on error?

sync_check_terraform/modules/sync_check/service/init.sh
```diff
@@ -0,0 +1,15 @@
...
+dnf install -y dnf-plugins-core
+dnf config-manager --add-repo https://download.docker.com/linux/fedora/docker-ce.repo
+dnf install -y docker-ce docker-ce-cli containerd.io
+dnf install -y docker-buildx-plugin docker-compose-plugin docker-compose
+dnf install -y ruby
+dnf install -y ruby-devel
+dnf install -y gcc make
+dnf clean all
+gem install slack-ruby-client
+gem install sys-filesystem
```
LesnyRumcajs, line 13: you can add multiple packages to install in one command, e.g., `dnf install ruby ruby devel gcc make ...`

LesnyRumcajs: same as in `gem`

LesnyRumcajs (commented):
@sudo-shashank Looks good. Could you please deploy it manually and run for a few days so we have confidence it can replace the former sync check service?

sync_check_terraform/modules/sync_check/service/run_service.sh
```diff
@@ -0,0 +1,74 @@
...
+docker volume create --name=forest-data
+docker volume create --name=sync-check
+docker volume create --name=ruby-common
+
+## We need it to access the DATA_DIR regardless of the user.
+chmod 0777 /var/lib/docker/volumes/forest-data/_data
+
+## Copy all relevant scripts
+cp -R /root/* /var/lib/docker/volumes/sync-check/_data/
+cp -R /root/ruby_common/* /var/lib/docker/volumes/ruby-common/_data/
```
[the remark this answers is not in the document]
aatifsyed:
```suggestion
## Copy all relevant scripts
cp --recursive /root/* /var/lib/docker/volumes/sync-check/_data/
cp --recursive /root/ruby_common/* /var/lib/docker/volumes/ruby-common/_data/
```

or are you asking me to push a commit?

sync_check_terraform/modules/sync_check/service/run_service.sh
```diff
@@ -0,0 +1,74 @@
+#!/bin/bash
+
+set -e
+
+# Check if an environment variable is set. If it isn't, set error=1.
+check_env () {
+    A="                            ";
+    echo -n "${A:0:-${#1}} $1: "
```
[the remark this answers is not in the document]
sudo-shashank: updated

.github/workflows/deploy-sync-check.yml
LesnyRumcajs: I asked the question before the `comment` jobs were added. :)

LesnyRumcajs (approved):
It seems to work fine. Let's decommission the current instance and merge this.

PR #107, opened 2023-06-15
Move snapshot service to edge

### LINKED ISSUE #106: Move snapshot service to `edge` in preparation of new release.
**Issue summary**

The snapshot service uses the latest stable release of Forest. Soon, a backward incompatible release is coming out. In preparation of this, we need to switch the service from `latest` to `edge`. This requires changing a few flags and a few other things.

Acceptance criteria:
- [ ] Snapshot service works with the `edge` tag.
- [ ] Snapshot service only uploads compressed snapshots.

sudo-shashank:
**Summary of changes**
- Changed `forest-tag` value to `edge` from `latest`.

Closes #106 


LesnyRumcajs (changes requested):
@sudo-shashank can you confirm that this will work fine? After https://github.com/ChainSafe/forest/pull/2982, all exported snapshots are compressed to my understanding.

This would mean we will compress an already compressed snapshot.
https://github.com/ChainSafe/forest-iac/blob/force-apply-on-dispatch/daily_snapshot_terraform/modules/daily_snapshot/service/upload_snapshot.sh#L55-L58

Or am I missing something?

lemmih:
Yeah, `edge` is not backward compatible with `latest` so you need to change a few flags. You can deploy this version under a different Terraform key and test that it works.

LesnyRumcajs (approved):
https://github.com/ChainSafe/forest-iac/blob/main/daily_snapshot_terraform/modules/daily_snapshot/service/init.sh#L3

The `zstd` package should not be needed anymore.

lemmih (approved):

PR #110, opened 2023-06-15
Installing New Relic Agents on all Forest Infrastructure Services

samuelarogbonlo:
**Summary of changes**
- Install new relic agents on all forest services with ansible

-
-

### DIFF
--- a/.github/workflows/deploy-forest.yml
+++ b/.github/workflows/deploy-forest.yml
@@ -29,7 +29,7 @@ jobs:
           aws_access_key_id:  ${{ secrets.AWS_ACCESS_KEY_ID }}
           aws_secret_access_key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           working_directory: terraform/mainnet
-          chain: mainnet
+          environment: mainnet
 
   deploy-forest-calibnet-infrastruture-with-terraform:
     name: Deploy Forest Terraform Calibnet Infrastructure
@@ -46,4 +46,4 @@ jobs:
           aws_access_key_id:  ${{ secrets.AWS_ACCESS_KEY_ID }}
           aws_secret_access_key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           working_directory: terraform/calibnet
-          chain: calibnet
+          environment: calibnet
--- a/composite-action/terraform/action.yml
+++ b/composite-action/terraform/action.yml
@@ -1,29 +1,28 @@
-name: Deploy Forest Infrastructure on DigitalOcean
+name: Custom Composite action to deploy terraform resources
 
 description: |
-  This action deploys the Forest infrastructure to either the mainnet or calibnet
-  chain using Terraform.
+  This action deploys the Forest infarstructure with Terraform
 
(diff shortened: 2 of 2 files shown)


.github/workflows/deploy-sync-check.yml
```diff
@@ -14,95 +14,35 @@ jobs:
...
-      - name: Terraform Plan Status
-        if: steps.plan.outcome == 'failure'
-        run: exit 1
+          do_token: ${{ secrets.DO_TOKEN }}
+          aws_access_key_id:  ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws_secret_access_key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          ssh_private_key: ${{ secrets.SSH_PRIVATE_KEY }}
+          slack_token: ${{ secrets.SLACK_TOKEN }}
+          working_directory: sync_check_terraform/
+          chain: Sync Check Service
```
lemmih, line 32: Let's change `chain` to something else. It doesn't appear to be a chain at all.

ansible/install_newrelic_infra_agent.yml
```diff
@@ -0,0 +1,13 @@
...
+  hosts: all
+  roles:
+    - role: newrelic.newrelic_install
+  vars:
+    targets:
+      - infrastructure
+      - logs 
+  environment:
+    NEW_RELIC_API_KEY: "{{ lookup('env', 'NEW_RELIC_API_KEY') }}"
+    NEW_RELIC_ACCOUNT_ID: "3942575"
```
lemmih, line 12: Are the services automatically updated if we change this account ID in the future?

terraform/modules/forest_node/user-data.sh
```diff
@@ -31,8 +31,12 @@ echo "AllowUsers ${NEW_USER}" >> /etc/ssh/sshd_config
 
 systemctl restart sshd
 
+# Enable passwordless sudo for the new user. This allows the user to run sudo commands without being prompted for a password.
+echo "${NEW_USER} ALL=(ALL:ALL) NOPASSWD: ALL" | sudo tee /etc/sudoers.d/"${NEW_USER}"
+
 # Add new user to "sudo" and "docker" group so they can run docker commands and have general admin rights.
-usermod --append --groups sudo,docker "${NEW_USER}"
+usermod --append --groups docker "${NEW_USER}"
+
```
lemmih, line 39: Being in the `sudo` group isn't enough?

ansible/install_newrelic_infra_agent.yml
lemmih: Can you explain to me how that works?

terraform/modules/forest_node/user-data.sh
lemmih: Could you elaborate in more detail? Why is being in the `sudo` group no longer sufficient?

lemmih: Shouldn't the NewRelic agent be installed by `root` rather than the `chainsafe` user?

PR #117, opened 2023-06-20
chore: Refactored to utilise  Terraform for installing the New Relic Agent and configured the integration of Prometheus OpenMetrics with New Relic

### LINKED ISSUE #97: Monitor services
**Issue summary**

Here's an incomplete list of attributes of the Forest Infrastructure that we would like to monitor:

- Snapshot buckets: https://github.com/ChainSafe/forest-iac/issues/94
- That the snapshot-service droplet is running.
- That the snapshot-service volume isn't full.
- That the sync-check droplet is running.
- That the Forest nodes (both mainnet and calibnet) are running and their epoch is advancing.
- That the Forest node volume isn't full.

samuelarogbonlo:
**Summary of changes**
- Refactored the code to utilize Terraform for installing the New Relic Agent instead of Ansible.
- Configured Prometheus OpenMetrics integration with New Relic for monitoring Forest Prometheus metrics.
- Developed a workaround for the lack of support in New Relic for Fedora machines, enabling the installation of the New Relic Agent. 

Closes #97 

### DIFF
--- a/terraform/modules/forest_node/user-data.sh
+++ b/terraform/modules/forest_node/user-data.sh
@@ -89,7 +89,7 @@ sudo --user="${NEW_USER}" -- \
   --volume=/var/run/docker.sock:/var/run/docker.sock \
   --restart=unless-stopped \
   containrrr/watchtower \
-  --label-enable --include-stopped --revive-stopped --stop-timeout 120s --interval 600
+  --include-stopped --revive-stopped --stop-timeout 120s --interval 600
 
 # Set-up  New Relic Agent For logs collection and Infrastruture Metrics
 sudo --user="${NEW_USER}" -- \


daily_snapshot_terraform/modules/daily_snapshot/service/init.sh
```diff
@@ -24,3 +24,24 @@ chmod 777 forest_db
...
+
+# Add the New Relic infrastructure monitoring agent repository to the system's list of yum repos.
+# The resulting file (/etc/yum.repos.d/newrelic-infra.repo) is used by yum to locate the New Relic packages.
+sudo curl -o /etc/yum.repos.d/newrelic-infra.repo https://download.newrelic.com/infrastructure_agent/linux/yum/el/9/x86_64/newrelic-infra.repo
+
+# Refreshes the New Relic repository. This step ensures that yum is aware of the latest versions of available packages.
+sudo yum -q makecache -y --disablerepo='*' --enablerepo='newrelic-infra'
+
+# Installs the New Relic infrastructure agent. This package provides the monitoring functionality needed.
+sudo yum install newrelic-infra -y
```
LesnyRumcajs, line 47: `yum` should not be used anymore in Fedora. If there's a compelling reason to use `yum` instead of `dnf`, it must be documented.

sync_check_terraform/modules/sync_check/service/init.sh
```diff
@@ -10,3 +10,24 @@ dnf clean all
...
+enable_process_metrics: true
+status_server_enabled: true
+status_server_port: 18003
+license_key: "$NR_LICENSE_KEY"
+custom_attributes:
+  nr_deployed_by: newrelic-cli
+display_name: "$NAME"
+EOF
+
+# Add the New Relic infrastructure monitoring agent repository to the system's list of yum repos.
```
LesnyRumcajs, line 25: yum should not be used anymore in Fedora. If there's a compelling reason to use yum instead of dnf, it must be documented.

terraform/mainnet/variable.tf
```diff
@@ -62,3 +62,17 @@ variable "fw_name" {
...
+}
+
+variable "NEW_RELIC_API_KEY" {
+  description = "The New Relic API KEY"
+  type        = string
+}
+
+
+
+
```
LesnyRumcajs, line 78:
```suggestion
```

terraform/modules/forest_node/user-data.sh
```diff
@@ -55,28 +59,73 @@ cat << EOF > "/home/${NEW_USER}/forest_data/config.toml"
...
   --restart=unless-stopped \
   containrrr/watchtower \
   --label-enable --include-stopped --revive-stopped --stop-timeout 120s --interval 600
+
+# Set-up  New Relic Agent For logs collection and Infrastruture Metrics
+sudo --user="${NEW_USER}" -- \
+  bash -c "curl -Ls https://download.newrelic.com/install/newrelic-cli/scripts/install.sh | bash && \
+  sudo NEW_RELIC_API_KEY=""${NEW_RELIC_API_KEY}"" \
+       NEW_RELIC_ACCOUNT_ID=""${NEW_RELIC_ACCOUNT_ID}"" \
+       NEW_RELIC_REGION=EU \
```
LesnyRumcajs, line 99: Should the NR region be parametrised?

terraform/modules/forest_node/variable.tf
```diff
@@ -70,3 +70,20 @@ variable "fw_name" {
...
+
+variable "NEW_RELIC_API_KEY" {
+  description = "New Relic API KEY"
+  type        = string
+}
+
+variable "NEW_RELIC_ACCOUNT_ID" {
+  description = "New Relic Account ID"
+  type        = string
+  default     = "3942575"
```
LesnyRumcajs, line 87: Is it our account id?

LesnyRumcajs (changes requested):

.github/workflows/deploy-forest-calibnet.yml
```diff
@@ -1,4 +1,4 @@
-name: Forest Calibnet
+name: Deploy
```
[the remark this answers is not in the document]
LesnyRumcajs: The name of the workflow should point to which service it touches. The name of the job should be the actual action being performed. So, in this case, I would expect `Forest Calibnet` as the workflow name and `Deploy` as the name for the `deploy-forest-calibnet-infrastruture-with-terraform`.

Do you agree @samuelarogbonlo @lemmih?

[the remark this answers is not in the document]
lemmih: As long as the names are short, I'm happy.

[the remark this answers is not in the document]
lemmih: `Forest Calibnet / Deploy` looks good to me.

terraform/modules/forest_node/variable.tf
LesnyRumcajs: Why do we expose our account id?

terraform/modules/forest_node/user-data.sh
LesnyRumcajs: I'd say this should be parametrised. If it is not, someone changing the node to NYC may be surprised he can't without changing a seemingly random shell script.

.github/workflows/deploy-sync-check.yml
```diff
@@ -30,3 +30,4 @@ jobs:
           slack_token: ${{ secrets.SLACK_TOKEN }}
           working_directory: sync_check_terraform/
           environment: Sync Check Service
+          NR_LICENSE_KEY: ${{ secrets.NR_LICENSE_KEY }}
```
ruseinov, line 33: nit: need to add a new line at EOF. Would recommend installing a linter going forward, easier to fix and verify.

sync_check_terraform/modules/sync_check/service/init.sh
```diff
@@ -10,3 +10,26 @@ dnf clean all
 gem install slack-ruby-client sys-filesystem
 
 nohup /bin/bash ./run_service.sh > run_service_log.txt &
+
```
ruseinov, line 13: All of this looks exactly the same as `daily_snapshot_terraform/modules/daily_snapshot/service/init.sh`. Which begs a question: should we have one template instead of these two files?

ruseinov:
I'm curious why aren't we using newrelic provider: https://docs.newrelic.com/docs/more-integrations/terraform/terraform-intro/

terraform/modules/forest_node/variable.tf
LesnyRumcajs: 1. The account may change for any reason. The ID should be easily changeable.
2. I don't know NR's policy on hiding the details of accounts (and no policies can protect accounts from leaks). It's best to not share even such things as an ID for privacy reasons.

LesnyRumcajs (approved):

PR #127, opened 2023-06-28
Run lotus nodes

### LINKED ISSUE #97: Monitor services

samuelarogbonlo:
**Summary of changes**
- Run the Lotus Mainnet and Calibnet node 
- Install the New Infra Agent in the nodes

Closes #97 

### DIFF
--- a/terraform/lotus-mainnet/main.tf
+++ b/terraform/lotus-mainnet/main.tf
@@ -3,7 +3,7 @@ terraform {
 
   backend "s3" {
     bucket                      = "forest-iac"
-    key                         = "lotus-mainner/terraform.tfstate"
+    key                         = "lotus-mainnet/terraform.tfstate"
     region                      = "us-west-1"
     endpoint                    = "fra1.digitaloceanspaces.com"
     skip_credentials_validation = true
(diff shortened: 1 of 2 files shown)


terraform/lotus-mainnet/main.tf
```diff
@@ -0,0 +1,35 @@
...
+    key                         = "lotus-mainner/terraform.tfstate"
+    region                      = "us-west-1"
+    endpoint                    = "fra1.digitaloceanspaces.com"
+    skip_credentials_validation = true
+    skip_metadata_api_check     = true
+  }
+}
+
+module "lotus-calibnet" {
+  source = "../modules/forest_node"
```
lemmih, line 15: I still need clarification. If the module is not specific to Forest, let's rename it. Perhaps `filecoin_node`.

lemmih (changes requested):

lemmih (approved):

PR #165, opened 2023-07-21
feat: docker-compose.yml for actor bundle ipfs file host

hanabi1224:
**Summary of changes**

Part of https://github.com/ChainSafe/forest/issues/2765

- docker-compose for actor bundle ipfs file host

The container is currently running on a DO droplet with `docker compose up --build --force-recreate --detach`, providing data to the IPFS public HTTP gateway.

Questions:
1. Do we want to use the private HTTP gateway? 
2. If the answer to 1 is yes, do we want to terraform this?
3. If the answer to 1 is yes, do we want to add SSL to the gateway (via ngnix and certbot)?

Example mapping: 
https://github.com/filecoin-project/builtin-actors/releases/download/v9.0.3/builtin-actors-calibrationnet.car at  https://ipfs.io/ipfs/QmVy2u2Gwa21hmbQvv7LGkGNTn2GmVxv1gwTfBeuWSBp1H

Closes 

### DIFF
--- a/ipfs/docker-compose.yml
+++ b/ipfs/docker-compose.yml
@@ -1,0 +1,23 @@
+# Docker compose file to run a ipfs node that serves the files defined in init.sh
+# Run command: `docker compose up --build --force-recreate --detach`
+# Check stats: `docker exec forest-ipfs ipfs stats bw`
+
+version: "3.7"
+
+services:
+  forest_ipfs:
+    image: ipfs/kubo
+    container_name: forest-ipfs
+    ports:
+      - "4001:4001"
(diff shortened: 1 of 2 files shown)


lemmih (changes requested):
We don't want to run more services than absolutely necessary. Let's discuss other solutions in the parent issue.

LesnyRumcajs:
> We don't want to run more services than absolutely necessary. Let's discuss other solutions in the parent issue.

That being said, using the tools from the Filecoin domain is very interesting. Sadly, at the moment they are not as reliable and performant as centralised solutions.

PR #173, opened 2023-07-23
fix: disable GCing when exporting a new snapshot

lemmih:
**Summary of changes**
- Locally, I get disconnects and `ChainExchange` failures when the GC runs. Maybe the GC blocks DB access which causes failed p2p responses which leads to disconnecting peers.

Closes 


LesnyRumcajs (approved):
Should it be disabled also during import? Or should even the GC task not start before the import is completed in Forest code directly?

lemmih:
> Should it be disabled also during import? Or should even the GC task not start before the import is completed in Forest code directly?

I think the GC only starts _after_ data has been imported. Not sure, thought.

PR #185, opened 2023-07-31
chore: droplet rename

samuelarogbonlo:
**Summary of changes**
- Rename forest nodes

Closes 


terraform/modules/filecoin_node/variable.tf
```diff
@@ -4,7 +4,7 @@ variable "image" {
 }
 
 variable "name" {
-  description = "The name of Forest Droplet"
+  description = "The name of Forest Droplet User"
```
lemmih, line 7: Name of the droplet user? Wasn't it right before?

lemmih (approved):

terraform/modules/filecoin_node/variable.tf
lemmih: What's a droplet user? Is this variable also used for the username on the droplet?

lemmih (approved):

PR #188, opened 2023-08-02
chore: deploy snapshot service without volume

### LINKED ISSUE #187: snapshot service: don't use volume
**Issue summary**

The DO Volumes are quite a bit slower than local storage. In the past, we had to use volumes because our database could grow bigger than the 320GiB available locally. However, we're now at a point where we'll never use more than 200GiB (database + any snapshots), and we can safely switch to using the local SSD.

Acceptance criteria:
- [x] Deploy the snapshot service _without_ a volume.

samuelarogbonlo:
**Summary of changes**
- Deploy the snapshot service without a volume.

Closes #187 

### DIFF
--- a/daily_snapshot_terraform/modules/daily_snapshot/service/upload_snapshot.sh
+++ b/daily_snapshot_terraform/modules/daily_snapshot/service/upload_snapshot.sh
@@ -61,10 +61,10 @@ write_metrics &
 
 forest-cli --config config.toml --chain "$CHAIN_NAME" db clean --force
 forest --config config.toml --chain "$CHAIN_NAME" --import-snapshot "$NEWEST_SNAPSHOT" --halt-after-import
-forest --config config.toml --chain "$CHAIN_NAME" --no-gc --detach
+forest --config config.toml --chain "$CHAIN_NAME" --no-gc --save-token=token.txt --detach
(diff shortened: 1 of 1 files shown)


daily_snapshot_terraform/modules/daily_snapshot/service/init.sh
```diff
@@ -18,9 +18,8 @@ s3cmd --dump-config \
     --secret_key="$AWS_SECRET_ACCESS_KEY" \
     --multipart-chunk-size-mb=4096 > ~/.s3cfg
 
-## Setup volume
+## Create forest data directory
```
LesnyRumcajs, line 21: this comment is not very helpful
```suggestion
```

daily_snapshot_terraform/modules/daily_snapshot/service/init.sh
```diff
@@ -7,7 +7,7 @@ gem install docker-api slack-ruby-client activesupport
 systemctl start docker
 
 # 1. Configure s3cmd
-# 2. Mount volume at 'forest_db'
+# 2. create forest_db directory
```
LesnyRumcajs, line 10: I don't think it's worth mentioning such trivialities

LesnyRumcajs (commented):
Have you verified that it works correctly?

lemmih (approved):

PR #189, opened 2023-08-02
fix: basic validation of snapshots before they're uploaded

lemmih:
**Summary of changes**
- Check that snapshots are at least somewhat valid before uploading them.

Closes 


daily_snapshot_terraform/modules/daily_snapshot/service/upload_snapshot.sh
```diff
@@ -64,6 +64,7 @@ forest --config config.toml --chain "$CHAIN_NAME" --import-snapshot "$NEWEST_SNA
 forest --config config.toml --chain "$CHAIN_NAME" --no-gc --detach
 timeout "$SYNC_TIMEOUT" forest-cli --chain "$CHAIN_NAME" sync wait
 forest-cli --chain "$CHAIN_NAME" snapshot export -o forest_db/
+forest-cli archive info forest_db/*.car.zst
```
LesnyRumcajs, line 67: What is the difference, in practice, between this command and `snapshot validate`?

lemmih: `archive info` doesn't check for bad links or execute any tipsets. It'll only fail if the snapshot is completely broken.

LesnyRumcajs (approved):

PR #190, opened 2023-08-02
chore: disable metrics logging

lemmih:
**Summary of changes**
- Logging metrics isn't all that useful anymore, and it makes the log files difficult to read.

Closes 

### DIFF
--- a/daily_snapshot_terraform/modules/daily_snapshot/service/upload_snapshot.sh
+++ b/daily_snapshot_terraform/modules/daily_snapshot/service/upload_snapshot.sh
@@ -57,7 +57,8 @@ echo "Chain: $CHAIN_NAME"
 echo "Snapshot: $NEWEST_SNAPSHOT"
 
 # spawn a task in the background to periodically write Prometheus metrics to a file
-write_metrics &
+# Disabled to reduce logging noise. Re-enable if Forest starts having network troubles again.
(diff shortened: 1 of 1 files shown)


daily_snapshot_terraform/modules/daily_snapshot/service/upload_snapshot.sh
```diff
@@ -57,7 +57,8 @@ echo "Chain: $CHAIN_NAME"
 echo "Snapshot: $NEWEST_SNAPSHOT"
 
 # spawn a task in the background to periodically write Prometheus metrics to a file
-write_metrics &
+# Disabled to reduce logging noise. Re-enable if Forest starts having network troubles again.
+# write_metrics &
```
LesnyRumcajs, line 61: They may be helpful. The noise comes from the fact that we have the debug option set with `x` that logs every single operation. I'd rather we remove this and not metrics.

lemmih: Hmm, right. Let's take a step back and re-think how we track the different logs.

LesnyRumcajs: related https://github.com/ChainSafe/forest-iac/issues/197

PR #204, opened 2023-08-04
chore: changed the base image from fedora to ubuntu

### LINKED ISSUE #166: Consider changing base images to more stable ones
**Issue summary**

While Fedora on its own is quite stable, we should limit our work related to maintenance. Fedora releases happen every six months, and they get one year of maintenance updates. This forces us to do updates at least yearly.

We could either go with Debian or [Ubuntu](https://wiki.ubuntu.com/Releases).

samuelarogbonlo:
**Summary of changes**
-  Changed the base images of Sync check and Snapshot services from fedora to ubuntu

Closes #166

### DIFF
--- a/terraform/modules/daily_snapshot/main.tf
+++ b/terraform/modules/daily_snapshot/main.tf
@@ -66,7 +66,7 @@ locals {
     "echo 'export SNAPSHOT_BUCKET=\"${var.snapshot_bucket}\"' >> .forest_env",
     "echo 'export SNAPSHOT_ENDPOINT=\"${var.snapshot_endpoint}\"' >> .forest_env",
     "echo 'export NEW_RELIC_API_KEY=\"${var.NEW_RELIC_API_KEY}\"' >> .forest_env",
-    "echo 'export NEW_RELIC_ACCOUNT_ID=\"${var.NEW_RELIC_ACCOUNT_ID}\"' .forest_env",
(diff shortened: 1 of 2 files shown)


terraform/modules/daily_snapshot/main.tf
```diff
@@ -65,7 +65,9 @@ locals {
     "echo 'export SLACK_NOTIF_CHANNEL=\"${var.slack_channel}\"' >> .forest_env",
     "echo 'export SNAPSHOT_BUCKET=\"${var.snapshot_bucket}\"' >> .forest_env",
     "echo 'export SNAPSHOT_ENDPOINT=\"${var.snapshot_endpoint}\"' >> .forest_env",
-    "echo 'export NR_LICENSE_KEY=\"${var.NR_LICENSE_KEY}\"' >> .forest_env",
+    "echo 'export NEW_RELIC_API_KEY=\"${var.NEW_RELIC_API_KEY}\"' >> .forest_env",
```
ruseinov, line 68: Curious why don’t we have a template instead of those “echo” scripts.

ruseinov: Yeah, so we can just interpolate the variables into it instead of echoing them one by one. 
I would consider doing as much as possible that was as opposed to embedding sh/bash 
into terraform configs.

ruseinov (approved):

terraform/modules/daily_snapshot/main.tf
ruseinov: Let's add a follow-up issue.

PR #214, opened 2023-08-15
fix: limit diff depth and use canonical names for diff snapshots

lemmih:
**Summary of changes**
- Diff snapshots may not omit data from _before_ the previous lite snapshot. Fixing this means using the `--diff-depth` flag.
- Include dates in the filenames of diff snapshots.

Closes 

### DIFF
--- a/archival-snapshots-generate/src/main.rs
+++ b/archival-snapshots-generate/src/main.rs
@@ -4,10 +4,14 @@
 use std::path::{Path, PathBuf};
 
 use anyhow::bail;
+use chrono::NaiveDateTime;
 use clap::Parser;
 use env_logger::Env;
 use log::{debug, info};
 use rayon::prelude::*;
+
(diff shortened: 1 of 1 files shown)


aatifsyed (approved):

elmattic (approved):

archival-snapshots-generate/src/main.rs
```diff
@@ -175,3 +188,15 @@ fn generate_diff_snapshot(
     info!("Generated diff snapshot for epochs: [{diff}, {epoch}]");
     Ok(())
 }
+
+fn epoch_to_date(epoch: ChainEpoch) -> String {
```
LesnyRumcajs, line 192: This is `mainnet` specific, no? Can we handle the calibnet case as well?

LesnyRumcajs: fixed

PR #215, opened 2023-08-15
fix: added more strict config to new relic

samuelarogbonlo:
**Summary of changes**
- Added more strict config to new relic

Closes 

### DIFF
--- a/terraform/modules/filecoin_node/forest.sh
+++ b/terraform/modules/filecoin_node/forest.sh
@@ -54,6 +54,7 @@ sudo --user="${NEW_USER}" -- \
   --detach \
   --network=forest \
   --name=forest-"${CHAIN}" \
+  --env "FOREST_GC_TRIGGER_FACTOR=1.4" \
   --volume=/home/"${NEW_USER}"/forest_data:/home/"${NEW_USER}"/forest_data:z \
   --publish=1234:1234 \
   --publish=6116:6116 \


terraform/modules/daily_snapshot/service/newrelic_fail2ban.sh
```diff
@@ -22,13 +24,15 @@ include_matching_metrics:
...
+metrics_process_sample_rate:-1
 metrics_system_sample_rate: 300
 metrics_storage_sample_rate: 300
+metrics_nfs_sample_rate: 300
+container_cache_metadata_limit: 300
 disable_zero_mem_process_filter: true
 disable_all_plugins: true
 disable_cloud_metadata: true 
-ignore_system_proxy: true 
+ignore_system_proxy: true
```
LesnyRumcajs, line 35:
```suggestion
ignore_system_proxy: true
```

terraform/modules/daily_snapshot/service/newrelic_fail2ban.sh
```diff
@@ -22,13 +24,15 @@ include_matching_metrics:
     - regex "^syslog.*"
     - regex "^gpg-agent.*"
 metrics_network_sample_rate: -1
-metrics_process_sample_rate: 300
+metrics_process_sample_rate:-1
```
LesnyRumcajs, line 27: I don't quite understand. This means we are disabling grabbing process metrics, right? Why? How much data does it generate? If it's really a lot, why do we still need `include_matching_metrics`?

terraform/modules/filecoin_node/forest.sh
```diff
@@ -85,18 +85,15 @@ if [ -n "${NEW_RELIC_API_KEY}" ] ; then
...
-    - regex "^forest.*"
-    - regex "^fail2ban.*"
-    - regex "^rsyslog.*"
-    - regex "^syslog.*"
-    - regex "^gpg-agent.*"
 metrics_network_sample_rate: -1
-metrics_process_sample_rate: 300
-metrics_system_sample_rate: 300
-metrics_storage_sample_rate: 300
+metrics_process_sample_rate: -1
```
LesnyRumcajs, line 92: Why do we discard Forest process metrics?

LesnyRumcajs: Please elaborate. I don't see how metric of 1 process, scraped every 5 minutes or so, are able to produce several gigabytes of data in a month.

LesnyRumcajs: What kind of data? I'd like us to understand this. I'm reluctant to believe that Linux generates so much data for a single process.

LesnyRumcajs (approved):

Issue #137, opened 2023-07-10
Benchmark improvements


lemmih:
**Issue summary**

- [x] Peak memory usage. Can be measured with `/usr/bin/time -v`.
- [ ] Validation time in a recent snapshot. Lotus doesn't have this functionality but we can fairly easily add it.
- [ ] Time to migrate network versions.
- [ ] Time to export a snapshot. Lotus requires more than 64GiB of memory to export a mainnet snapshot. Find a way to set a memory limit. If that's too difficult then we can just measure calibnet export.
- [ ] Time to compute the state of a single tipset (using `lotus state compute-state`).
- [ ] Use `lotus-shed` to benchmark individual functions without running the lotus-daemon.
- [ ] Use published Docker images instead of building Lotus on our own https://github.com/ChainSafe/forest-iac/issues/143
- [ ] [Create common error handler](https://github.com/ChainSafe/forest/pull/2367#discussion_r1195390109).
- [ ] [Check preconditions before running script (enough disk space and `ulimit -n` is correctly set).](https://github.com/ChainSafe/forest/pull/2367#discussion_r1200295926)
- [ ] [Delete cloned git repos at the end of script instead of after each client run.](https://github.com/ChainSafe/forest/pull/2367#discussion_r1200311414) (not applicable if/when the task to use published Docker images above is completed)
- [ ] [Investigate and fix issue with missing gems when using `bundle install`.](https://github.com/ChainSafe/forest/pull/2367#issuecomment-1558977266)
- [ ] Other misc improvements/optimizations (e.g., [long form command line options](https://github.com/ChainSafe/forest/pull/2367#discussion_r1196605001), [modify `private` module designations](https://github.com/ChainSafe/forest/pull/2367#discussion_r1195387339), [replace erroneous references to `heights` with `epochs` or `tipsets`](https://github.com/ChainSafe/forest/issues/2925)).
- [ ] Ensure snapshot is deleted (only if downloaded) after a successful run (we only seem to be deleting a downloaded snapshot in case of an error).

**Other information and links**

Migrated tasks from analogous [issue in Forest repo](https://github.com/ChainSafe/forest/issues/2683).

jdjaustin:
Currently working on the task to create a common error handler and simultaneously improving error handling (currently we suppress detailed errors).

LesnyRumcajs:
The CPU usage should also be added to the result of each action.

elmattic:
[...]

PR #122, opened 2023-06-22
Add a new benchmark service

sudo-shashank:
**Summary of changes**
Changes introduced in this pull request:

Added a new benchmark service:
- Continuously runs benchmarks 
```
ruby bench.rb --chain calibnet --tempdir ./snapshots --daily
ruby bench.rb --chain mainnet --tempdir ./snapshots --daily
```
- Upload benchmark results to [`forest-benchmark`](https://cloud.digitalocean.com/spaces/forest-benchmarks?i=88c522) [weekly-results](https://cloud.digitalocean.com/spaces/forest-benchmarks?i=88c522&path=benchmark-results%2Fweekly-results%2F) and [all_results](https://forest-benchmarks.fra1.digitaloceanspaces.com/benchmark-results/all_results.csv) file.
- Sends regular status updates on slack `#forest-notifications` channel.

**Reference issue to close (if applicable)**

Closes #92 

**Other information and links**

PR #228, opened 2023-08-23
[Refactor] Benchmark script error handling improvements

jdjaustin:
**Summary of changes**
- Implement improvements to error handling. Create a common error handler and use this error handler where possible; also propagate detailed errors.

Closes associated tasks in #137 

Based off of changes in #122; will wait for that PR to merge before merging this PR.


terraform/modules/benchmark/service/bench.rb
```diff
@@ -270,10 +267,12 @@ def run_benchmarks(benchmarks, options)
   Dir.chdir(WORKING_DIR) do
     benchmarks_loop(benchmarks, options, bench_metrics)
   end
-  if options[:daily]
-    write_csv(bench_metrics, options)
-  else
-    write_markdown(bench_metrics)
+  handle_exception(nil) do
```
elmattic, line 270: Using `false` instead of `nil` could be a bit clearer.

elmattic: Or even `@db_created = false`.

terraform/modules/benchmark/service/benchmark_base.rb
```diff
@@ -1,5 +1,26 @@
 # frozen_string_literal: true
 
+def cleanup(db_created)
+  # Delete downloaded snapshot if it exists.
+  FileUtils.rm_f(@snapshot_path) unless !@snapshot_downloaded || @snapshot_path.nil?
+  FileUtils.rm_rf("#{WORKING_DIR}/snapshot_dl_files")
```
elmattic, line 6: Probably out of scope for this PR, but are we doing this kind of cleanup in case of a successful run?

terraform/modules/benchmark/service/benchmark_base.rb
```diff
@@ -166,20 +187,16 @@ def run_validation_step(daily, args, metrics)
 
   # Import snapshot, write metrics, and call validation function, returning metrics.
   def import_and_validation(daily, args, metrics)
-    import_command = splice_args(@import_command, args)
-    metrics[:import] = exec_command(import_command)
+    handle_exception(true) do
```
elmattic, line 190: I do understand the nested call to `handle_exception` but maybe we could force deletion of `db` in all cases?

terraform/modules/benchmark/service/benchmark_base.rb
jdjaustin: Added this to #137

terraform/modules/benchmark/service/benchmark_base.rb
jdjaustin: Just documenting our latest in person discussion: the `--force` flag is already specified, but the `-c` flag fails if `handle_exception` is called before the `data_dir` is created, so we need to keep the nested call with a `db_created` argument to specify when the `data_dir` is created.

terraform/modules/benchmark/service/bench.rb
jdjaustin: Updated.

terraform/modules/benchmark/service/benchmark_base.rb
elmattic: Ok then maybe we could just test the presence of the config file, and if present call the `db destroy` whether the db is existing or not (if db is not found, the command still returns a success).

PR #230, opened 2023-08-28
chore: use env template file for setting secret

### LINKED ISSUE #208: Consider using templates for setting secret variables
**Issue` summary**

https://github.com/ChainSafe/forest-iac/blob/f6678397046ac7d28edc2790293f2581f007e65f/terraform/modules/sync_check/main.tf#L61-L68
we should consider using `env_templates` we can just interpolate the variables into it instead of echoing them one by one. Affected services are the `sync-check` and `snapshot`

samuelarogbonlo:
**Summary of changes**
- use env template for setting secret

Closes #208 

### DIFF
--- a/terraform/modules/sync_check/service/init.sh
+++ b/terraform/modules/sync_check/service/init.sh
@@ -6,7 +6,7 @@ set -eux
 
 # Use APT specific mechanism to ensure non-interactive operation and wait for the lock
 sudo DEBIAN_FRONTEND=noninteractive apt-get -qqq --yes -o DPkg::Lock::Timeout=-1 update
-sudo DEBIAN_FRONTEND=noninteractive apt-get -qqq --yes -o DPkg::Lock::Timeout=-1 install -y ruby ruby-dev s3cmd anacron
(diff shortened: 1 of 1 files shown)


terraform/modules/sync_check/service/init.sh
```diff
@@ -6,7 +6,7 @@ set -eux
 
 # Use APT specific mechanism to ensure non-interactive operation and wait for the lock
 sudo DEBIAN_FRONTEND=noninteractive apt-get -qqq --yes -o DPkg::Lock::Timeout=-1 update
-sudo DEBIAN_FRONTEND=noninteractive apt-get -qqq --yes -o DPkg::Lock::Timeout=-1 install -y ruby ruby-dev s3cmd anacron
+sudo DEBIAN_FRONTEND=noninteractive apt-get -qqq --yes -o DPkg::Lock::Timeout=-1 install -y ruby ruby-dev gcc make
```
LesnyRumcajs, line 9: Please explain why a template requires `gcc` and `make`.

LesnyRumcajs: So what does it have to do with templates?

LesnyRumcajs: If nothing, this should be a separate PR.

LesnyRumcajs (approved):

PR #257, opened 2023-09-11
feat: Measure Peak Memory Usage 

samuelarogbonlo:
**Summary of changes**
- Measure Peak Memory Usage 

part of #137

Closes

### DIFF
--- a/terraform/modules/benchmark/service/benchmark_base.rb
+++ b/terraform/modules/benchmark/service/benchmark_base.rb
@@ -38,7 +38,6 @@ module ExecCommands
   def measure_online_validation(benchmark, pid, metrics)
     Thread.new do
       first_epoch = nil
-
       loop do
         start, first_epoch = benchmark.start_online_validation_command
         if start
@@ -47,23 +46,12 @@ module ExecCommands
         end
         sleep 0.1
       end
-
       sleep benchmark.online_validation_secs
-
       last_epoch = get_last_epoch(benchmark)
       metrics[:num_epochs] = last_epoch - first_epoch
-      @logger.info 'Stopping process...'
 
-      # Identify and stop the child process
-      child_pid = `pgrep -P #{pid}`.strip.to_i
-      if child_pid.zero?
-        @logger.error "Failed to find child PID for parent PID: #{pid}"
-        benchmark.stop_command(pid)
-      else
-        @logger.info "Identified child PID: #{child_pid}"
-        benchmark.stop_command(pid)
-        benchmark.stop_command(child_pid)
-      end
+      @logger.info 'Stopping process...'
+      benchmark.stop_command(pid)
     end
   end
 


terraform/modules/benchmark/service/benchmark_base.rb
```diff
@@ -15,8 +15,11 @@ def get_last_epoch(benchmark)
 
   # Measures validation time for daily metrics.
   def measure_online_validation(benchmark, pid, metrics)
+    # Start a new thread
```
LesnyRumcajs, line 18: How is this comment helpful?

terraform/modules/benchmark/service/bench.rb
```diff
@@ -149,15 +149,22 @@ def write_csv(metrics, options)
...
+      validation_time: { forest: 'n/a', lotus: 'n/a' },
+      peak_memory: { forest: 'n/a', lotus: 'n/a' }
+    }
 
     metrics.each do |key, value|
       elapsed = value[:import][:elapsed] || 'n/a'
       tpm = value[:validate_online][:tpm] || 'n/a'
+      peak_memory_import = value[:import][:peak_memory] || 'n/a'
+      peak_memory_validate = value[:validate_online][:peak_memory] || 'n/a'
+      peak_memory = [peak_memory_import, peak_memory_validate].max
```
LesnyRumcajs, line 163: This will throw an exception if classes of `peak_memory_import` and `peak_memory_validate` do not match.

```
irb(main):005:0> ['n/a', 3].max
(irb):5:in `<main>': comparison of Integer with String failed (ArgumentError)
```

terraform/modules/benchmark/service/benchmark_base.rb
```diff
@@ -25,15 +28,34 @@ def measure_online_validation(benchmark, pid, metrics)
         end
         sleep 0.1
       end
+
+      # Sleep for the specified validation duration
```
LesnyRumcajs, line 32: How is this comment helpful?

terraform/modules/benchmark/service/benchmark_base.rb
```diff
@@ -25,15 +28,34 @@ def measure_online_validation(benchmark, pid, metrics)
...
       sleep benchmark.online_validation_secs
+
+      # Calculate the number of epochs
       last_epoch = get_last_epoch(benchmark)
       metrics[:num_epochs] = last_epoch - first_epoch
-
       @logger.info 'Stopping process...'
-      benchmark.stop_command(pid)
+
+      # Identify and stop the child process
```
LesnyRumcajs, line 40: Perhaps this should be a part of another PR. What does it have to do with measuring peak memory usage?

terraform/modules/benchmark/service/benchmark_base.rb
```diff
@@ -58,19 +80,26 @@ def proc_monitor(pid, benchmark)
   end
 
   # Helper function for measuring execution time; passes process ID to online
-  # validation and process monitor.
+  # validation and process monitor and Measure Peak Memory.
```
LesnyRumcajs, line 83: I don't understand this.

terraform/modules/benchmark/service/benchmark_base.rb
```diff
@@ -58,19 +80,26 @@ def proc_monitor(pid, benchmark)
...
   def exec_command_aux(command, metrics, benchmark)
-    Open3.popen2(*command) do |i, o, t|
+    command_with_time = ['/usr/bin/time', '-v'] + command
```
LesnyRumcajs, line 85: won't `['/usr/bin/time', '-v', command]` work?

terraform/modules/benchmark/service/benchmark_base.rb
```diff
@@ -58,19 +80,26 @@ def proc_monitor(pid, benchmark)
...
+
+    Open3.popen2e(*command_with_time) do |i, o_and_err, t|
       pid = t.pid
       i.close
 
+      # Start the process monitor immediately after starting the process, as in the original.
```
LesnyRumcajs, line 91: what original?

terraform/modules/benchmark/service/benchmark_base.rb
```diff
@@ -58,19 +80,26 @@ def proc_monitor(pid, benchmark)
...
+      output = []
+      o_and_err.each_line do |l|
+        output << l.force_encoding('UTF-8')
         print l
       end
 
-      handle.join
-      metrics.merge!(proc_metrics)
+      # Extract the memory usage and update the metrics.
+      metrics[:peak_memory] = extract_memory_usage(output.join("\n"))
```
LesnyRumcajs, line 100: you don't need double quotes for literals.

LesnyRumcajs (changes requested):

terraform/modules/benchmark/service/benchmark_base.rb
LesnyRumcajs: indeed, my bad.

terraform/modules/benchmark/service/benchmark_base.rb
LesnyRumcajs: why?

terraform/modules/benchmark/service/benchmark_base.rb
LesnyRumcajs: I don't quite understand. Do you mean that stopping the process of `/usr/bin/time <some-command>` does not stop its child processes? They are left orphaned?

LesnyRumcajs: I don't think this is the right way to handle this, then. There are standard UNIX tools to manage killing a parent with their children. Either use the group PID to kill the parent (it should then kill the children as well, as they share this identifier) or alternatively there's a `-P` flag for `pkill` (some logic would need to be changed for this, to my understanding of the script)

terraform/modules/benchmark/service/benchmark_base.rb
LesnyRumcajs: Do you know why?

LesnyRumcajs: So how can you fix this?

LesnyRumcajs: I'd rather have something more explicit (the `command` being array is a bit unexpected), but I'm okay with leaving this as is.

LesnyRumcajs (approved):

Issue #247, opened 2023-09-06
Benchmark service should only notify the Slack channel once per day


lemmih:
**Issue summary**

The benchmark service notifies the slack channel after each successful run. This should be changed to only notify once per day.

**Other information and links**

lemmih:
> @lemmih just to clarify, should it just run once per day or find a logic for it to notify only once per day but still run continuously as it is now

It should run continuously but only report to Slack once per day (or even less frequently than that).

LesnyRumcajs:
How about a summary of benchmark runs from the past day? It would nicely show how results changed, and we could act on this if they get weird.

lemmih:
Including useful information in the status report would be a nice feature but it is out of scope for this issue. This issue is merely about not spamming the Slack channel.

LesnyRumcajs:
@samuelarogbonlo to make it easy, let's just follow the daily snapshot steps.

https://github.com/ChainSafe/forest-iac/blob/f3b5b418b3725a2f463026ad71592750b98effea/terraform/modules/daily_snapshot/service/daily_snapshot.rb#L41-L44

Once per day for now is fine. Later we can increase the interval. Successful operation definition remains the same as for normal run - the process exits with code 0. That is to say, we care only about the benchmark run at hand.

lemmih:
> @lemmih okay i understand that, But what I'm still trying to grasp is: to prevent spamming, should the success message be sent only once a day or week? And what, in your opinion, should define a successful operation since the benchmark runs consistently? That part is a bit unclear to me.

The current service sends a message to #forest-notifications after every successful run and after every failed run. We _only_ want to change how often success messages are sent to the Slack channel. This does not require re-defining what a successful run is.

PR #265, opened 2023-09-19
fix: ensure benchmark service notify once per day

samuelarogbonlo:
**Summary of changes**
- ensure benchmark service notify once per day

first implementation  #247

Closes

### DIFF
--- a/cf/latest-snapshot/.gitignore
+++ b/cf/latest-snapshot/.gitignore
@@ -1,0 +1,1 @@
+.wrangler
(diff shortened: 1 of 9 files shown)


LesnyRumcajs (commented):
It would be more robust to check the modification date of the benchmark results (or, even better, the last entry there). This would ensure that the notifications are not sent on service restart (which is the case with this proposal).

What do you think? We can merge this but continue in a separate PR (just don't close the original issue)

LesnyRumcajs (approved):

Issue #264, opened 2023-09-19
Support range queries when fetching snapshots


lemmih:
**Issue summary**

We currently ignore range queries when downloading snapshots from CloudFlare. Without range queries, we cannot resume downloads.

Tasks:
- [ ] Set up local development environment with `wrangler dev`.
- [ ] Download the latest calibnet snapshot from your local deployment. (This will use the `forest-archive-dev` bucket rather than the production `forest-archive` bucket).
- [ ] Confirm that range queries do not work.
- [ ] Extract range parameters from the HTTP request (see: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Range).
- [ ] Pass range parameters to R2.get (see: https://developers.cloudflare.com/r2/api/workers/workers-api-reference/#ranged-reads)
- [ ] Confirm that resume works.
- [ ] Confirm that downloading multiple chunks in parallel works (use `aria2c -x5`).

This is the `R2.get` method that needs range options:
https://github.com/ChainSafe/forest-iac/blob/4a2d495e5397ac1347f14db8d0cb48ed948d408b/cf/latest-snapshot/src/index.ts#L19

**Other information and links**

R2 worker API: https://developers.cloudflare.com/r2/api/workers/workers-api-reference/

PR #266, opened 2023-09-20
CF latest snapshot: support range queries

LesnyRumcajs:
**Summary of changes**
- added range queries support to the latest snapshot worker
- tested with `aria2c -x5` and Brave browser that resumes work correctly,
- enforce `GET` method for all fetches

Closes https://github.com/ChainSafe/forest-iac/issues/264

### DIFF
--- a/cf/latest-snapshot/src/index.ts
+++ b/cf/latest-snapshot/src/index.ts
@@ -36,10 +36,17 @@ async function get_latest(req_headers: Headers, env: Env, chain: string): Promis
 		let status = 200;
 		let range = parseRange(object.size, req_headers.get("range") || "");
 		if (Array.isArray(range)) {
+			// R2Object doesn't support multiple ranges, so we only use the first one.
+			// Throw an error if there are more than one range.
+			if (range.length > 1) {
+				return new Response('Multiple ranges are not supported', {
+					status: 416, // Range Not Satisfiable
+				});
+			}
+
+			let r = range[0];
+			headers.set("Content-Range", `${r.type} ${r.start}-${r.end}/${object.size}`);
 			status = 206; // Partial Content
-			range.forEach(function (r) {
-				headers.append("Content-Range", `${range.type} ${r.start}-${r.end}/${object.size}`);
-			})
 		}
 
 		object.writeHttpMetadata(headers);
--- a/cf/latest-snapshot/src/index.ts
+++ b/cf/latest-snapshot/src/index.ts
@@ -1,3 +1,5 @@
+import parseRange from 'range-parser';
+
 interface Env {
 	FOREST_ARCHIVE: R2Bucket,
 }
@@ -7,7 +9,7 @@ function basename(path: String) {
 }
 
 // Directly fetch the data for the latest snapshot of a given chain (eg. calibnet or mainnet)
-async function get_latest(env: Env, chain: string): Promise<Response> {
+async function get_latest(req_headers: Headers, env: Env, chain: string): Promise<Response> {
 	const listed = await env.FOREST_ARCHIVE.list({ prefix: chain + "/latest/" });
 	let latest = listed.objects.at(-1);
 	if (latest == null) {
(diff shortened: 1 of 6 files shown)


cf/latest-snapshot/src/index.ts
```diff
@@ -7,31 +9,56 @@ function basename(path: String) {
...
 	let latest = listed.objects.at(-1);
 	if (latest == null) {
 		return new Response(`No latest snapshot found ${chain}`, {
 			status: 404,
 		});
 	} else {
-		// Should we support range queries?
-		const object = await env.FOREST_ARCHIVE.get(latest.key);
+		const object = await env.FOREST_ARCHIVE.get(latest.key, {
+			range: req_headers,
```
lemmih, line 21: Interesting. I didn't know you could pass in the headers directly for the `range` field. The documentation indicates that you cannot.

cf/latest-snapshot/src/index.ts
```diff
@@ -40,6 +67,16 @@ export default {
 		const url = new URL(request.url);
 		const chain = (url.pathname.match(/\/latest\/(\w*)/) || ["undefined"])[1];
 
-		return await get_latest(env, chain);
+		// Disallow any other request method except GET, they are not sensible in the context
+		// of fetching a snapshot.
+		if (request.method !== 'GET') {
```
lemmih, line 72: Should we support `HEAD`? How does `aria2c` figure out how large the files are?

cf/latest-snapshot/src/index.ts
```diff
@@ -7,31 +9,56 @@ function basename(path: String) {
...
-		let encoded_name = encodeURIComponent(basename(object.key));
+		const encoded_name = encodeURIComponent(basename(object.key));
 		// Tell browsers and aria2c which filename to use. For 'wget', you have to use `--trust-server-names`.
-		headers.set('content-disposition', `attachment; filename*=UTF-8''${encoded_name}; filename="${encoded_name}"`);
+		headers.set('Content-Disposition', `attachment; filename*=UTF-8''${encoded_name}; filename="${encoded_name}"`);
 
-		return new Response(object.body, {
-			headers,
-		});
+		if ('body' in object) {
```
lemmih, line 51: When does `object` not have `body`?

cf/latest-snapshot/src/index.ts
LesnyRumcajs: What documentation are you referring to? The API seems to be happy with it.
```
  get(
    key: string,
    options: R2GetOptions & {
      onlyIf: R2Conditional | Headers;
    }
  ): Promise<R2ObjectBody | R2Object | null>;
```
...
```
declare interface R2GetOptions {
  onlyIf?: R2Conditional | Headers;
  range?: R2Range | Headers;
}
```

cf/latest-snapshot/src/index.ts
LesnyRumcajs: I have no clue under what conditions the `get` would return an `R2Object` (and not an `R2ObjectBody`). It is a possible output from this method, though, so it needs to be handled (otherwise, we would have a runtime error).

cf/latest-snapshot/src/index.ts
```diff
@@ -7,31 +9,56 @@ function basename(path: String) {
...
 			});
 		}
+
 		const headers = new Headers();
+
+		// If the client requested a range, then we need to set the Content-Range header.
+		// The `parseRange` returns an error code if the range is not satisfiable (or not specified),
+		// so we can handle the range request only if it returns an array.
+		let status = 200;
+		let range = parseRange(object.size, req_headers.get("range") || "");
```
lemmih, line 37: Why not use the range from `object.range`?

cf/latest-snapshot/src/index.ts
lemmih: It always returns both, doesn't it?

lemmih: The returned object implements both `R2Object` and `R2ObjectBody`.

LesnyRumcajs: Does it implement _both_, though? Or does it implement _either_? I got warnings that `R2Object` does not have a `body` property.

LesnyRumcajs: That is to say, I do believe it's a union (in C sense) where only one variant is valid.

cf/latest-snapshot/src/index.ts
LesnyRumcajs: The function accepts strings, and not `R2Range`.

cf/latest-snapshot/src/index.ts
LesnyRumcajs: `aria2c -x5` seems to be happy without it. It calls GET only.

```
12:03:32 PM GET /latest/calibnet/ 200
12:03:33 PM GET /latest/calibnet/ 206
12:03:33 PM GET /latest/calibnet/ 206
12:03:33 PM GET /latest/calibnet/ 206
12:03:33 PM GET /latest/calibnet/ 206
```

cf/latest-snapshot/src/index.ts
LesnyRumcajs: Unless you are referring to replacing `parseRange` with an `R2Range` altogether, which would bloat the code a bit, given how `R2Rrange` is implemented:
```
declare type R2Range =
  | {
      offset: number;
      length?: number;
    }
  | {
      offset?: number;
      length: number;
    }
  | {
      suffix: number;
    };
```

lemmih: Wouldn't it be simpler than what you have now? Anyhow, I'll leave it in your capable hands.

LesnyRumcajs: Sadly, I don't think it will be simpler, and it would potentially introduce some corner cases (length larger than the total size, suffix larger than the total size...)

cf/latest-snapshot/src/index.ts
```diff
@@ -7,31 +9,63 @@ function basename(path: String) {
...
+		const encoded_name = encodeURIComponent(basename(object.key));
 		// Tell browsers and aria2c which filename to use. For 'wget', you have to use `--trust-server-names`.
-		headers.set('content-disposition', `attachment; filename*=UTF-8''${encoded_name}; filename="${encoded_name}"`);
+		headers.set('Content-Disposition', `attachment; filename*=UTF-8''${encoded_name}; filename="${encoded_name}"`);
 
-		return new Response(object.body, {
-			headers,
-		});
+		if ('body' in object) {
+			return new Response(object.body, {
```
lemmih, line 59: I feel there should be an easier way of doing this but I know no JS.

lemmih (approved):

cf/latest-snapshot/src/index.ts
LesnyRumcajs: I found no easier way (though many more difficult and esoteric ways which strangely resembling C++ template metaprogramming)

PR #285, opened 2023-10-02
fix: segregate daily_snapshot logs

samuelarogbonlo:
**Summary of changes**
- Output from the daemon, error messages from the daemon, script execution, logs from background-running daemons, and metrics are each saved in different files. In the event of a failure, each file is individually attached to the Slack thread

Closes 

### DIFF
--- a/.github/workflows/deploy-daily-snapshot.yml
+++ b/.github/workflows/deploy-daily-snapshot.yml
@@ -8,6 +8,9 @@ on:
     paths:
       - 'terraform/daily_snapshot/**'
       - 'terraform/modules/daily_snapshot/**'
+    # This needs to be declared explicitly so that the job is actually
+    # run when moved out of draft.
+    types: [opened, synchronize, reopened, ready_for_review]
   push:
     branches:
       - main
--- a/composite-action/terraform/action.yml
+++ b/composite-action/terraform/action.yml
@@ -77,17 +77,20 @@ runs:
         TF_VAR_NEW_RELIC_ACCOUNT_ID: ${{ inputs.new_relic_account_id }}
 
     - name: Find Comment
-      if: github.event_name == 'pull_request'
+      if: github.event.pull_request.draft == false &&
+          github.event_name == 'pull_request'
       uses: peter-evans/find-comment@v2
       id: fc
       with:
         issue-number: ${{ github.event.pull_request.number }}
         comment-author: 'github-actions[bot]'
-        body-regex: "^### Forest ${{ inputs.environment }} Infrastructure Plan"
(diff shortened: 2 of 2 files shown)


terraform/modules/daily_snapshot/service/upload_snapshot.sh
```diff
@@ -25,18 +27,26 @@ apt-get update && apt-get install -y curl
...
+    {
+      curl --silent --fail --max-time 5 --retry 5 --retry-delay 2 --retry-max-time 10 http://localhost:6116/metrics || true
+    } | add_timestamps >> "/home/$LOG_EXPORT_METRICS"
     sleep 5
   done
 }
 
 function print_forest_logs {
-  cat forest.err forest.out metrics.txt
+  cat forest.err forest.out > /home/$LOG_EXPORT_DAEMON
```
lemmih, line 49: Why is `/home/` here?

terraform/modules/daily_snapshot/service/upload_snapshot.sh
```diff
@@ -89,6 +100,7 @@ docker run \
   --rm \
   --user root \
   -v "$CHAIN_DB_DIR:/home/forest/forest_db":z \
+  -v "$CHAIN_LOGS_DIR:/home/logs":z \
```
lemmih, line 103: Could you explain the naming? `/home/logs` looks strange to me.

terraform/modules/daily_snapshot/service/upload_snapshot.sh
lemmih: Why are the logs in the home dir?

terraform/modules/daily_snapshot/service/upload_snapshot.sh
lemmih: The `/home/` folder is for user accounts. `logs` is not a user.

terraform/modules/daily_snapshot/service/upload_snapshot.sh
```diff
@@ -94,6 +105,7 @@ docker run \
   --rm \
   --user root \
   -v "$CHAIN_DB_DIR:/home/forest/forest_db":z \
+  -v "$CHAIN_LOGS_DIR:$CHAIN_LOGS_DIR":z \
```
lemmih, line 108: Using fixed mount points inside the container is preferable to avoid accidental clashes. How about putting the logs at `/home/forest/logs`?

lemmih (approved):
This adds quite a bit of complexity, but maybe there's no way around that.

LesnyRumcajs (approved):

PR #289, opened 2023-10-04
improvement: use cheaper and faster droplet

lemmih:
**Summary of changes**
- the `s-4vcpu-16gb-amd` droplet is both cheaper and faster than `so-2vcpu-16gb` at exporting. Probably because the SSD is faster. This change should make snapshot exports 10-20 minutes faster.

Closes

### DIFF
--- a/terraform/daily_snapshot/main.tf
+++ b/terraform/daily_snapshot/main.tf
@@ -29,7 +29,7 @@ module "daily_snapshot" {
 
   # Configure service:
   name              = "forest-snapshot"       # droplet name
-  size              = "so-2vcpu-16gb"         # droplet size
+  size              = "s-4vcpu-16gb-amd"      # droplet size
   slack_channel     = "#forest-notifications" # slack channel for notifications
   snapshot_bucket   = "forest-snapshots"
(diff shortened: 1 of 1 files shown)


elmattic (approved):

terraform/daily_snapshot/main.tf
```diff
@@ -29,7 +29,7 @@ module "daily_snapshot" {
 
   # Configure service:
   name              = "forest-snapshot"       # droplet name
-  size              = "so-2vcpu-16gb"         # droplet size
+  size              = "s-4vcpu-16gb-amd"      # droplet size
```
LesnyRumcajs, line 32: Shall we make the exact change in the sync check? Or anywhere else?

lemmih: `s-4vcpu-16gb-amd` is faster but only has 200GiB of storage. It should be enough for the sync check unless you're doing strange things with the snapshots. :)

LesnyRumcajs: In the past 14 days, we never exceeded 40% of the 300 GB disk, so we should be good for the sync check.
image

PR #303, opened 2023-10-06
fix: wait for cloud-init to finish

lemmih:
**Summary of changes**
- Cloud-init clashes with our init script. We should wait for the machine to finish initializing before running our own script.

Closes

### DIFF
--- a/terraform/modules/daily_snapshot/service/init.sh
+++ b/terraform/modules/daily_snapshot/service/init.sh
@@ -2,12 +2,15 @@
 
 set -eux
 
+# Wait for cloud-init to finish initializing the machine
+cloud-init status --wait
+
 # Setting DEBIAN_FRONTEND to ensure non-interactive operations for APT
 export DEBIAN_FRONTEND=noninteractive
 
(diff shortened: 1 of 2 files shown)


elmattic (approved):

LesnyRumcajs (commented):
Shall we do the same for the sync check?

lemmih:
> Shall we do the same for the sync check?

Yes, I think it'll solve a lot of the random failures we've been seeing. We can probably also remove the silly sleep at the end of the initialization.

LesnyRumcajs:
> > Shall we do the same for the sync check?
> 
> Yes, I think it'll solve a lot of the random failures we've been seeing. We can probably also remove the silly sleep at the end of the initialization.

Let's do it, then.

lemmih:
> > > Shall we do the same for the sync check?
> > 
> > 
> > Yes, I think it'll solve a lot of the random failures we've been seeing. We can probably also remove the silly sleep at the end of the initialization.
> 
> Let's do it, then.

Done.

LesnyRumcajs (approved):

PR #307, opened 2023-10-06
chore: generate at most 1 mainnet snapshot per hour

lemmih:
**Summary of changes**
- Turns out we don't need that many mainnet snapshots. One per hour is enough.

Closes


ruseinov (approved):

LesnyRumcajs (approved):
this snippet gets more action then our lockfiles

Issue #267, opened 2023-09-20
Testing for Cloudflare workers


LesnyRumcajs:
**Issue summary**

The [CF workers] are not covered by any tests, making any modification to them risky. Investigate the best approach to introduce testing to them (might be some TypeScript technology, might be just plain shell scripts) and implement the tests.

**Other information and links**

PR #330, opened 2023-11-03
Cloudflare CI/CD

LesnyRumcajs:
**Summary of changes**
- CI/CD for CF workers. It doesn't do any solid testing (https://github.com/ChainSafe/forest-iac/issues/267), but with the `--dry-run`, we are at least able to detect something like, e.g., broken `wrangler.toml`.

Closes

### DIFF
--- a/.github/workflows/deploy-cf-prune-latest.yml
+++ b/.github/workflows/deploy-cf-prune-latest.yml
@@ -1,0 +1,28 @@
+name: Deploy CF prune-latest
+on:
+  pull_request:
+    paths:
+      - 'cf/prune-latest'
+      - '.github/workflows/deploy-cf-prune-latest.yml'
+  push:
+    paths:
+      - 'cf/prune-latest'
+      - '.github/workflows/deploy-cf-prune-latest.yml'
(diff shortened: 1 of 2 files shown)


lemmih (approved):

lemmih (approved):

.github/workflows/deploy-cf-prune-latest.yml
```diff
@@ -0,0 +1,28 @@
+name: Deploy CF prune-latest
+on:
+  pull_request:
+    paths:
+      - 'cf/prune-latest'
+      - '.github/workflows/deploy-cf-prune-latest.yml'
+  push:
+    paths:
```
hanabi1224, line 8: Maybe also limit the branch?
```yaml
push:
  branches:
        - main
   path:
```

LesnyRumcajs: Why would we limit it to `main`? We want the `--dry-check` to run on pushes to branches as well.

hanabi1224 (approved):

PR #341, opened 2023-11-12
downgrade snapshot service to v0.15.2

LesnyRumcajs:
**Summary of changes**
- the `v0.16.0` _Rottenmelon_ may have a performance regression. For the time being, we could use `v0.15.2` _Defenestration_, given we are past the Watermelon fixes epochs anyway with the recent snapshots.

Closes

### DIFF
--- a/terraform/daily_snapshot/prod/main.tf
+++ b/terraform/daily_snapshot/prod/main.tf
@@ -36,7 +36,7 @@ module "daily_snapshot" {
   size            = "s-4vcpu-16gb-amd"      # droplet size
   slack_channel   = "#forest-notifications" # slack channel for notifications
   snapshot_bucket = "forest-archive"
-  forest_tag      = "v0.16.0"
+  forest_tag      = "v0.15.2"
   r2_endpoint     = "https://2238a825c5aca59233eab1f221f7aefb.r2.cloudflarestorage.com/"
 
   # Variable passthrough:


lemmih (approved):

lemmih (changes requested):
The snapshot server has synced, and a snapshot is being exported. Let's hold off this PR for now.

LesnyRumcajs:
I did a single check of ~80 epochs between `0.16.0` and `0.15.2`, and there was a 20% difference (`0.16.0` being slower) with a bunch of `sending on a closed channel` warnings. It's not decisive, though, the `0.16.0` may have gotten better peers.

LesnyRumcajs:
@lemmih as confirmed by @ruseinov, https://github.com/ChainSafe/forest/pull/3683 caused the regression in performance so I believe we should go ahead with the downgrade for now.

ruseinov:
@LesnyRumcajs perhaps instead of reverting we should do a hotfix to `0.16`? 
The problem is we already have some snapshots that don't work with `0.15.2` version..

LesnyRumcajs:
It shouldn't matter really. The snapshot service is bootstrapping from newer snapshots anyway.

ruseinov:
> It shouldn't matter really. The snapshot service is bootstrapping from newer snapshots anyway.

Roger that. 
I had issues with bootstrapping `0.15.2` with the newest calibnet snapshot.

ruseinov:
We're holding off on this PR according to @lemmih, not sure if we need to merge it then or not.

LesnyRumcajs:
Hopefully, superseded by https://github.com/ChainSafe/forest-iac/pull/342

PR #344, opened 2023-11-16
feat: mirror filecoin built-in actors

### LINKED ISSUE #323: Mirror builtin-actors releases
**Issue summary**

Given that Forest relies on `builtin-actors` releases, we want to ensure higher availability in case GH is down. The idea is to mirror the [builtin-actors releases](https://github.com/filecoin-project/builtin-actors/releases/) with a GHA cron job.

Sample implementation:
1. Create a workflow running every hour.
2. Check if we need to do any mirroring (by comparing lists of assets if possible)
3. If yes, upload the missing assets.

Perhaps there are some off-the-shelf solutions we can re-use. Or perhaps we could use Gitea to mirror (and host!) releases automatically?

samuelarogbonlo:
**Summary of changes**
- added a workflow running every hour.
- Downloads filecoin built-in actors released in the last two years
- Checks if we need to do any mirroring by comparing downloaded actors byte for byte with the ones in the object storage(R2 and DO)
- If yes, upload the missing assets.

Closes #323 

### DIFF
--- a/.github/workflows/deploy-snapshot-monitoring.yml
+++ b/.github/workflows/deploy-snapshot-monitoring.yml
@@ -1,0 +1,50 @@
+name: Snapshot Monitoring Service
+concurrency: ci-${{ github.ref }}-snapshot-monitoring
+
+on:
+  pull_request:
+    branches:
+      - main
+    paths:
+      - 'tf-managed/modules/snapshot-monitoring/**'
+      - 'tf-managed/live/environments/prod/applications/snapshot-monitoring/**'
+    # This needs to be declared explicitly so that the job is actually
+    # run when moved out of draft.
+    types: [opened, synchronize, reopened, ready_for_review]
+  push:
+    branches:
+      - main
+    paths:
+      - 'tf-managed/modules/snapshot-monitoring/**'
+      - 'tf-managed/live/environments/prod/applications/snapshot-monitoring/**'
+  workflow_dispatch:
+
+jobs:
+  deploy-snapshot-monitoring:
+    env:
+      TF_VAR_slack_enable: true
+      TF_VAR_slack_destination_id: ${{ secrets.SLACK_DESTINATION_ID }}
+      TF_VAR_slack_channel_id: ${{ secrets.SLACK_CHANNEL_ID }}
+      TF_VAR_pagerduty_enable: true
+      TF_VAR_pagerduty_destination_id: ${{ secrets.PAGERDUTY_DESTINATION_ID }}
+    runs-on: ubuntu-latest
+    permissions: write-all
+    steps:
+      - name: Checkout the code
+        uses: actions/checkout@v4
+
+      # Using Custom Composite action in ./composite-action/terragrunt folder
+      - name: Composite Action for Deploying Terragrunt Resources
+        uses: ./composite-action/terragrunt
+        with:
+          do_token: ${{ secrets.DO_TOKEN }}
+          aws_access_key_id:  ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws_secret_access_key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          r2_access_key: ${{ secrets.R2_ACCESS_KEY }}
+          r2_secret_key: ${{ secrets.R2_SECRET_KEY }}
+          slack_token: ${{ secrets.SLACK_TOKEN }}
+          working_directory: tf-managed/live/environments/prod/applications/snapshot-monitoring
+          service_name: Snapshot Service
+          new_relic_account_id: ${{ secrets.NEW_RELIC_ACCOUNT_ID }}
+          new_relic_api_key: ${{ secrets.NEW_RELIC_API_KEY }}
+          ssh_private_key: ${{ secrets.SSH_PRIVATE_KEY }}
--- a/tf-managed/modules/snapshot-monitoring/README.md
+++ b/tf-managed/modules/snapshot-monitoring/README.md
@@ -1,0 +1,18 @@
+# Snapshot monitoring
+
+This module creates New Relic resources to check the status of the global snapshot service. Note that this does not check the status of the [snapshot-service](../daily-snapshot), but rather the actual snapshots (and the epochs at which they were produced), available at:
+- <https://forest-archive.chainsafe.dev/mainnet/latest/>
+- <https://forest-archive.chainsafe.dev/calibnet/latest/>
+
+The logic is contained in the [snapshot-age-monitor](./snapshot-age-monitor.js).
+
+Networks: `mainnet` and `calibnet`
+Checks:
+- asserts the epoch of the hosted snapshots is not older than a given threshold,
+- trigger alarms if more than 2/3 checks over an hour fail.
+
+## Alarm destinations
+
+All alarms are handled via NewRelic, which integrates other applications. For now, it's:
+* [PagerDuty](https://chainsafe.pagerduty.com/) - the provided destination is linked with a service generated there and visible in the [service direcory](https://chainsafe.pagerduty.com/service-directory). It will fire alarms to on-call engineers if enabled, so use it cautiously.
+* [Slack](https://api.slack.com/) - the provided destination is also linked with company Slack, and the `channel_id` governs the actual channel to which the notifications will go.
--- a/tf-managed/modules/snapshot-monitoring/main.tf
+++ b/tf-managed/modules/snapshot-monitoring/main.tf
@@ -1,3 +1,11 @@
+locals {
+  service_name = format("%s-snapshot-age-monitor", var.environment)
+}
+
+resource "newrelic_alert_policy" "alert" {
+  name = format("%s alert policy", local.service_name)
+}
+
 resource "newrelic_synthetics_script_monitor" "snapshot-age-monitor" {
   status = "ENABLED"
   name   = format("%s-snapshot-age-monitor", var.environment)
@@ -14,6 +22,125 @@ resource "newrelic_synthetics_script_monitor" "snapshot-age-monitor" {
 
   tag {
     key    = "service"
-    values = ["forest-snapshot"]
+    values = ["forest-snapshot", var.environment]
+  }
+}
+
+resource "newrelic_notification_channel" "slack-channel" {
+  count          = var.slack_enable ? 1 : 0
+  name           = format("%s slack", local.service_name)
+  type           = "SLACK"
+  destination_id = var.slack_destination_id
+  product        = "IINT"
+
+  property {
+    key   = "channelId"
+    value = var.slack_channel_id
+  }
+
+  property {
+    key   = "customDetailsSlack"
+    value = "issue id - {{issueId}}"
+  }
+}
+
+resource "newrelic_notification_channel" "pagerduty-channel" {
+  count          = var.pagerduty_enable ? 1 : 0
+  name           = format("%s pagerduty", local.service_name)
+  type           = "PAGERDUTY_SERVICE_INTEGRATION"
+  destination_id = var.pagerduty_destination_id
+  product        = "IINT"
+
+  property {
+    key   = "summary"
+    value = "Filecoin Forest snapshot age monitor {{#isCritical}}CRITICAL{{/isCritical}}{{#isWarning}}WARNING{{/isWarning}}"
+  }
+
+  property {
+    key   = "customDetails"
+    value = <<-EOT
+            {
+            "id":{{json issueId}},
+            "IssueURL":{{json issuePageUrl}},
+            "NewRelic priority":{{json priority}},
+            "Total Incidents":{{json totalIncidents}},
+            "Impacted Entities":"{{#each entitiesData.names}}{{this}}{{#unless @last}}, {{/unless}}{{/each}}",
+            "Runbook":"{{#each accumulations.runbookUrl}}{{this}}{{#unless @last}}, {{/unless}}{{/each}}",
+            "Description":"{{#each annotations.description}}{{this}}{{#unless @last}}, {{/unless}}{{/each}}",
+            "isCorrelated":{{json isCorrelated}},
+            "Alert Policy Names":"{{#each accumulations.policyName}}{{this}}{{#unless @last}}, {{/unless}}{{/each}}",
+            "Alert Condition Names":"{{#each accumulations.conditionName}}{{this}}{{#unless @last}}, {{/unless}}{{/each}}",
+            "Workflow Name":{{json workflowName}}
+            }
+        EOT
+  }
+}
+
+resource "newrelic_nrql_alert_condition" "failing-snapshot-age" {
+  policy_id = newrelic_alert_policy.alert.id
+  name      = format("%s failing snapshot age", local.service_name)
+  enabled   = true
+  # Bound to the monitor query interval which is `EVERY_HOUR`
+  aggregation_window = 3600
+
+  nrql {
+    query = "SELECT filter(count(*), WHERE result = 'FAILED') AS 'Failures' FROM SyntheticCheck WHERE monitorName = '${local.service_name}'"
+  }
+
+  # This means that all locations are failing
+  critical {
+    operator              = "above_or_equals"
+    threshold             = 3
+    threshold_duration    = 3600
+    threshold_occurrences = "ALL"
+  }
+
+  warning {
+    operator              = "above_or_equals"
+    threshold             = 2
+    threshold_duration    = 3600
+    threshold_occurrences = "ALL"
+  }
+}
+
+resource "newrelic_workflow" "alerting-workflow-slack" {
+  count                 = var.slack_enable ? 1 : 0
+  name                  = format("%s slack alerting workflow", local.service_name)
+  muting_rules_handling = "NOTIFY_ALL_ISSUES"
+
+  issues_filter {
+    name = format("%s alerting workflow filter", local.service_name)
+    type = "FILTER"
+
+    predicate {
+      attribute = "labels.policyIds"
+      operator  = "EXACTLY_MATCHES"
+      values    = [newrelic_alert_policy.alert.id]
+    }
+  }
+
+  destination {
+    channel_id = newrelic_notification_channel.slack-channel[0].id
+  }
+}
+
+resource "newrelic_workflow" "alerting-workflow-pagerduty" {
+  count                 = var.pagerduty_enable ? 1 : 0
+  name                  = format("%s pagerduty alerting workflow", local.service_name)
+  muting_rules_handling = "NOTIFY_ALL_ISSUES"
+
+  issues_filter {
+    name = format("%s alerting workflow filter", local.service_name)
+    type = "FILTER"
+
+    predicate {
+      attribute = "labels.policyIds"
+      operator  = "EXACTLY_MATCHES"
+      values    = [newrelic_alert_policy.alert.id]
+    }
+  }
+
+  destination {
+    channel_id = newrelic_notification_channel.pagerduty-channel[0].id
   }
 }
--- a/tf-managed/modules/snapshot-monitoring/provider.tf
+++ b/tf-managed/modules/snapshot-monitoring/provider.tf
@@ -3,7 +3,7 @@ terraform {
   required_providers {
     newrelic = {
       source  = "newrelic/newrelic"
-      version = "~> 3.0"
+      version = "~> 3.5"
     }
   }
 }
--- a/tf-managed/modules/snapshot-monitoring/variable.tf
+++ b/tf-managed/modules/snapshot-monitoring/variable.tf
@@ -22,3 +22,43 @@ variable "new_relic_account_id" {
   type        = number
   sensitive   = true
 }
+
+variable "slack_enable" {
+  description = "Enable Slack notifications"
+  type        = bool
+  default     = false
+}
+
+# This needs to be created manually. Afterwards, it can be found in
+# NR / Alerts & AI / Destinations.
+variable "slack_destination_id" {
+  description = "Slack destination id"
+  type        = string
+  default     = ""
+  sensitive   = true
+}
+
+# Channel ID - due to the limitations of NR it's not a human readable name
+# but an ID. This can be found at the bottom of the channel settings window.
+variable "slack_channel_id" {
+  description = "Slack channel id"
+  type        = string
+  default     = ""
+  sensitive   = true
+}
+
+variable "pagerduty_enable" {
+  description = "Enable PagerDuty notifications"
+  type        = bool
+  default     = false
+}
+
+# This needs to be created manually. Afterwards, it can be found in
+# NR / Alerts & AI / Destinations. Otherwise, we'd need special permissions
+# to create it via the API.
+variable "pagerduty_destination_id" {
+  description = "PagerDuty destination id"
+  type        = string
+  default     = ""
+  sensitive   = true
+}


LesnyRumcajs:
Mirroring only the latest release is not useful. The goal is to mirror all releases, properly versioned. The end structure of the mirror should be like (though more complete):
```
├── v10.0.0
│   ├── builtin-actors-devnet.car
│   └── builtin-actors-mainnet.car
├── v10.0.0-rc.1
│   └── builtin-actors-calibrationnet.car
├── v11.0.0
│   ├── builtin-actors-devnet.car
│   └── builtin-actors-mainnet.car
```

LesnyRumcajs (changes requested):
Mirroring only the latest release is not useful. The goal is to mirror all releases, properly versioned. The end structure of the mirror should be like (though more complete):
```
├── v10.0.0
│   ├── builtin-actors-devnet.car
│   └── builtin-actors-mainnet.car
├── v10.0.0-rc.1
│   └── builtin-actors-calibrationnet.car
├── v11.0.0
│   ├── builtin-actors-devnet.car
│   └── builtin-actors-mainnet.car
```

samuelarogbonlo:
@LesnyRumcajs ping

LesnyRumcajs:
@samuelarogbonlo it still doesn't work, or am I missing something? I'm looking at https://cloud.digitalocean.com/spaces/filecoin-builtin-actors?path

There are no releases mirrored for, e.g., version 8 or 9. Also, there are some stray releases that shouldn't be there, like `fip0061` or `tmp`. We don't want them.

samuelarogbonlo:
> @samuelarogbonlo it still doesn't work, or am I missing something? I'm looking at https://cloud.digitalocean.com/spaces/filecoin-builtin-actors?path
> 
> There are no releases mirrored for, e.g., version 8 or 9. Also, there are some stray releases that shouldn't be there, like `fip0061` or `tmp`. We don't want them.

@LesnyRumcajs i didn’t mirror from Genesis but I will do that now. For the releases we don't want can i get a list of them? or it's just `fip0061` and `tmp`

LesnyRumcajs:
> > @samuelarogbonlo it still doesn't work, or am I missing something? I'm looking at https://cloud.digitalocean.com/spaces/filecoin-builtin-actors?path
> > There are no releases mirrored for, e.g., version 8 or 9. Also, there are some stray releases that shouldn't be there, like `fip0061` or `tmp`. We don't want them.
> 
> @LesnyRumcajs i didn’t mirror from Genesis but I will do that now. For the releases we don't want can i get a list of them? or it's just `fip0061` and `tmp`

There's no such list; we need to be pragmatic. If it doesn't cost much, we can mirror everything. That'd be just a bit wasteful. For example, currently, Forest requires such bundles:
```
"v9.0.3" for "calibrationnet",
"v10.0.0-rc.1" for "calibrationnet",
"v11.0.0-rc2" for "calibrationnet",
"v12.0.0-rc.1" for "calibrationnet",
"v12.0.0-rc.2" for "calibrationnet",
"v12.0.0" for "calibrationnet",
"v11.0.0" for "butterflynet",
"v12.0.0" for "butterflynet",
"v9.0.3" for "devnet",
"v10.0.0" for "devnet",
"v11.0.0" for "devnet",
"v12.0.0" for "devnet",
"v9.0.3" for "mainnet",
"v10.0.0" for "mainnet",
"v11.0.0" for "mainnet",
"v12.0.0" for "mainnet",
```
And so, for the moment, we could grab whatever release matching `^v\d+\.\d+\.\d+.*$`. It'd be useful to generatee first such list for proofreading.

samuelarogbonlo:
> And so, for the moment, we could grab whatever release matching `^v\d+\.\d+\.\d+.*$`. It'd be useful to generatee first such list for proofreading.

```
Generated list for proofreading:
v12.0.0
v12.0.0-rc.3
v12.0.0-rc.2
v12.0.0-rc.1
v12.0.0-dev.3
v12.0.0-dev.2
v12.0.0-dev.1
v12.0.0-dev.0
v11.0.0
v11.0.0-rc2-alpha
v11.0.0-rc2
v11.0.0-rc1
v10.0.0
v10.0.0-rc.1
v9.0.3-wallaby
v8.0.0-wallaby
v9.0.3
v9.0.2
v9.0.1
v9.0.0
v9.0.0-rc.2
v9.0.0-rc.1
v8.0.0
```
@LesnyRumcajs

LesnyRumcajs:
The list looks complete and includes all of the releases we need; see https://github.com/ChainSafe/forest/blob/210d8c69fbd83b799a1a849ffcdab3d500fe7d7f/src/networks/actors_bundle.rs#L65-L80

Let's do the mirroring to CF; it's only on DO. After cleaning up the script, I'll review it. (there are some mentions of `LIST_FILE="$(pwd)/release_list_for_review.txt"` which shouldn't be in the service code).

samuelarogbonlo:
> The list looks complete and includes all of the releases we need; see https://github.com/ChainSafe/forest/blob/210d8c69fbd83b799a1a849ffcdab3d500fe7d7f/src/networks/actors_bundle.rs#L65-L80
> 
> Let's do the mirroring to CF; it's only on DO. After cleaning up the script, I'll review it. (there are some mentions of `LIST_FILE="$(pwd)/release_list_for_review.txt"` which shouldn't be in the service code).

@LesnyRumcajs done

.github/workflows/mirror-builtin-actors.yml
```diff
@@ -0,0 +1,61 @@
...
+          access_key: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          secret_key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+
+      - name: Run mirroring script to DigitalOcean
+        working-directory: scripts
+        if: github.ref == 'refs/heads/main' && (github.event_name == 'push' || github.event_name == 'workflow_dispatch')
+        run: bash ./mirror-builtin-actors.sh
+        env:
+            SLACK_API_TOKEN: ${{ secrets.SLACK_TOKEN }}
+            SLACK_CHANNEL: "#forest-dump"
```
LesnyRumcajs, line 34:
```suggestion
            SLACK_CHANNEL: "#forest-notifications"
```

.github/workflows/mirror-builtin-actors.yml
```diff
@@ -0,0 +1,61 @@
...
+        uses: actions/checkout@v2
+
+      - name: Set up S3cmd cli tool for Cloudflare
+        uses: s3-actions/s3cmd@v1.5.0
+        with:
+          provider: cloudflare
+          region: auto
+          access_key: ${{ secrets.R2_ACCESS_KEY }}
+          secret_key: ${{ secrets.R2_SECRET_KEY }}
+          account_id: "2238a825c5aca59233eab1f221f7aefb"
```
LesnyRumcajs, line 51: While not extremely sensitive, the ID should not be also hidden behind a secret.

.github/workflows/mirror-builtin-actors.yml
```diff
@@ -0,0 +1,61 @@
...
+
+      - name: Run script to mirror to Cloudflare
+        working-directory: scripts
+        if: github.ref == 'refs/heads/main' && (github.event_name == 'push' || github.event_name == 'workflow_dispatch')
+        run: bash ./mirror-builtin-actors.sh
+        env:
+            SLACK_API_TOKEN: ${{ secrets.SLACK_TOKEN }}
+            SLACK_CHANNEL: "#forest-dump"
```
LesnyRumcajs, line 59:
```suggestion
            SLACK_CHANNEL: "#forest-notifications"
```

scripts/mirror-builtin-actors.sh
```diff
@@ -0,0 +1,114 @@
...
+# - Sends an alert to Slack if any uploads fail.
+
+set -eo pipefail
+
+export DEBIAN_FRONTEND=noninteractive
+
+BASE_FOLDER="$(pwd)/releases/actors"
+API_URL="https://api.github.com/repos/filecoin-project/builtin-actors/releases"
+FAILED_LOG="$(pwd)/failed_uploads.log"
+TWO_YEARS_AGO=$(date -d '3 years ago' +%s)
```
LesnyRumcajs, line 18: The variable name is not consistent with the code.

scripts/mirror-builtin-actors.sh
LesnyRumcajs, line 1: Unless particularly obvious (`ls -alh`), let's use long flag names in scripts.

scripts/mirror-builtin-actors.sh
```diff
@@ -0,0 +1,114 @@
...
+# Function to extract the next page URL from GitHub API response headers for pagination.
+get_next_page_url() {
+    local headers="$1"
+    echo "$headers" | grep -oP '<\K([^>]+)(?=>; rel="next")' || echo ""
+}
+
+# Function to fetch and process releases
+fetch_and_process_releases() {
+    local api_url="$1"
+    local page_url="$api_url"
```
LesnyRumcajs, line 31: What's the point of the `api_url` variable?

scripts/mirror-builtin-actors.sh
```diff
@@ -0,0 +1,114 @@
...
+# - Uploads new or updated assets to the S3 bucket.
+# - Sends an alert to Slack if any uploads fail.
+
+set -eo pipefail
+
+export DEBIAN_FRONTEND=noninteractive
+
+BASE_FOLDER="$(pwd)/releases/actors"
+API_URL="https://api.github.com/repos/filecoin-project/builtin-actors/releases"
+FAILED_LOG="$(pwd)/failed_uploads.log"
```
LesnyRumcajs, line 17: This file is never read, and it is no longer available after the action is complete. What's its purpose?

scripts/mirror-builtin-actors.sh
```diff
@@ -0,0 +1,114 @@
...
+            popd > /dev/null
+        else
+            echo "No assets found for $TAG_NAME."
+        fi
+    fi
+done < <(find "$BASE_FOLDER" -mindepth 1 -type d)
+
+: 'Send summary alert only if there were failed uploads'
+if [ ${#failed_uploads[@]} -ne 0 ]; then
+    send_slack_alert_with_failed
```
LesnyRumcajs, line 111: Is this a function? Where is it declared?

scripts/mirror-builtin-actors.sh
```diff
@@ -0,0 +1,114 @@
...
+            echo "Processing assets for $TAG_NAME..."
+
+            echo "$ASSETS" | while IFS= read -r asset; do
+                DOWNLOAD_URL=$(echo "$asset" | jq -r '.browser_download_url')
+                FILE_NAME=$(echo "$asset" | jq -r '.name')
+
+                echo "Checking asset: $FILE_NAME"
+                if [ ! -f "$FILE_NAME" ]; then
+                    echo "Downloading $FILE_NAME..."
+                    wget -q "$DOWNLOAD_URL" -O "$FILE_NAME" || echo "Failed to download $FILE_NAME"
```
LesnyRumcajs, line 80: Sticking to either `curl` or `wget` would be more consistent.

.github/workflows/mirror-builtin-actors.yml
```diff
@@ -0,0 +1,61 @@
...
+            ENVIROMENT: cloudflare
```
LesnyRumcajs, line 60: there's a typo here, plus, I don't see where this variable is used. Could you point me to it?

scripts/mirror-builtin-actors.sh
```diff
@@ -0,0 +1,114 @@
...
+
+# This script mirrors all releases of Filecoin's builtin-actors that have been updated in the past two years,
+# It respects GitHub API rate limits and paginates requests.
+# It performs the following operations:
+# - Downloads release assets from GitHub.
+# - Compares these assets with the existing ones in an S3 bucket.
+# - Uploads new or updated assets to the S3 bucket.
+# - Sends an alert to Slack if any uploads fail.
+
+set -eo pipefail
```
LesnyRumcajs, line 11: We should also check for unset variables.

LesnyRumcajs (changes requested):

scripts/mirror-builtin-actors.sh
samuelarogbonlo: sure

.github/workflows/mirror-builtin-actors.yml
samuelarogbonlo: I had updated the script from using environment and didn't update

scripts/mirror-builtin-actors.sh
samuelarogbonlo: used to pass an input when running the function but i will update

scripts/mirror-builtin-actors.sh
samuelarogbonlo: missed it from playing around with the script, but updated it now

scripts/mirror-builtin-actors.sh
samuelarogbonlo: pushed the wrong last commit, it's read now

.github/workflows/mirror-builtin-actors.yml
samuelarogbonlo: ping @LesnyRumcajs

LesnyRumcajs: I forgot how to use English. The variable should not be exposed. It should be either a secret or a variable. Being paranoid, I'd opt for it being a secret so it is hidden in the logs.

LesnyRumcajs: Cloudflare account id is available under `CLOUDFLARE_ACCOUNT_ID` secret, see https://github.com/ChainSafe/forest-iac/blob/terragrunt-iac/.github/workflows/deploy-cf-prune-latest.yml#L27

scripts/mirror-builtin-actors.sh
```diff
@@ -0,0 +1,126 @@
...
+
+            echo "$ASSETS" | while IFS= read -r asset; do
+                DOWNLOAD_URL=$(echo "$asset" | jq --raw-output '.browser_download_url')
+                FILE_NAME=$(echo "$asset" | jq --raw-output '.name')
+
+                echo "Checking asset: $FILE_NAME"
+                if [ ! -f "$FILE_NAME" ]; then
+                    echo "Downloading $FILE_NAME..."
+                    curl --location --silent --show-error --output "$FILE_NAME" "$DOWNLOAD_URL" || echo "Failed to download $FILE_NAME"
+
```
LesnyRumcajs, line 92:
```suggestion
```

LesnyRumcajs (commented):
It'd be useful to verify checksums of the CAR files, given they are quite critical to be correct and we are given the expected checksums anyway. We can do it in a separate PR, though.

I invoke @aatifsyed, a world-class expert in `bash`-script technology. I find it a bit difficult to read.

aatifsyed (changes requested):
Congratulations on getting this bash script working :)

I think at this level of complexity, I'd rather see it in rust or even python, otherwise the maintenance burden will grow steeply.

e.g
```python
import os
from datetime import datetime, timezone

import boto3
import requests
from dateutil.relativedelta import relativedelta
from github import Github
from slack_sdk.web import WebClient

slack = WebClient(token=os.environ["SLACK_TOKEN"])

try:
    github = Github()
    s3 = boto3.client("s3")

    S3_BUCKET = "filecoin-builtin-actors"

    already_mirrored = set(obj["Name"] for obj in s3.list_objects(Bucket=S3_BUCKET))

    for release in github.get_repo("filecoin-project/builtin-actors").get_releases():
        age = abs(relativedelta(datetime.now(timezone.utc), release.published_at))
        if age.years > 3:
            continue  # skip old releases

        for asset in release.assets:
            key = f"{release.tag_name}/{asset.name}"
            response = requests.get(asset.url)
            response.raise_for_status()
            if key not in already_mirrored:
                # NOTE: this holds the file in memory, but that's expected to be small for these assets
                s3.put_object(Bucket=S3_BUCKET, Key=key, Body=response.content)

except Exception as e:
    slack.chat_postMessage(
        channel="#forest-notifications", text=f"Actor mirroring failed: {e}"
    ).validate()
```

We can chat about dependency management if you don't have a solution already

Also, I don't think every script should have to worry about failure - can we have a `job` that runs `on: failure` to report failing runs automatically?

scripts/mirror-builtin-actors.sh
```diff
@@ -0,0 +1,126 @@
...
+            echo "Processing assets for $TAG_NAME..."
+
+            echo "$ASSETS" | while IFS= read -r asset; do
+                DOWNLOAD_URL=$(echo "$asset" | jq --raw-output '.browser_download_url')
+                FILE_NAME=$(echo "$asset" | jq --raw-output '.name')
+
+                echo "Checking asset: $FILE_NAME"
+                if [ ! -f "$FILE_NAME" ]; then
+                    echo "Downloading $FILE_NAME..."
+                    curl --location --silent --show-error --output "$FILE_NAME" "$DOWNLOAD_URL" || echo "Failed to download $FILE_NAME"
```
aatifsyed, line 91: This silently continues the script on failure, no?

scripts/mirror-builtin-actors.sh
```diff
@@ -0,0 +1,126 @@
...
+
+                fi
+
+                echo "Checking $FILE_NAME against S3 version..."
+                TEMP_S3_DIR=$(mktemp --directory)
+                s3cmd get --no-progress "s3://$BUCKET_NAME/$TAG_NAME/$FILE_NAME" "$TEMP_S3_DIR/$FILE_NAME" || true
```
aatifsyed, line 97: why download the file here? these assets should be kept immutable

samuelarogbonlo: we're downloading it so that we can compare the s3 version against the one in Github. If there's any difference or missing actor, upload the downloaded actor from Github

samuelarogbonlo:
> Congratulations on getting this bash script working :)
> 
> I think at this level of complexity, I'd rather see it in rust or even python, otherwise the maintenance burden will grow steeply.
> 
> e.g
> 
> ```python
> import os
> from datetime import datetime, timezone
> 
> import boto3
> import requests
> from dateutil.relativedelta import relativedelta
> from github import Github
> from slack_sdk.web import WebClient
> 
> slack = WebClient(token=os.environ["SLACK_TOKEN"])
> 
> try:
>     github = Github()
>     s3 = boto3.client("s3")
> 
>     S3_BUCKET = "filecoin-builtin-actors"
> 
>     already_mirrored = set(obj["Name"] for obj in s3.list_objects(Bucket=S3_BUCKET))
> 
>     for release in github.get_repo("filecoin-project/builtin-actors").get_releases():
>         age = abs(relativedelta(datetime.now(timezone.utc), release.published_at))
>         if age.years > 3:
>             continue  # skip old releases
> 
>         for asset in release.assets:
>             key = f"{release.tag_name}/{asset.name}"
>             response = requests.get(asset.url)
>             response.raise_for_status()
>             if key not in already_mirrored:
>                 # NOTE: this holds the file in memory, but that's expected to be small for these assets
>                 s3.put_object(Bucket=S3_BUCKET, Key=key, Body=response.content)
> 
> except Exception as e:
>     slack.chat_postMessage(
>         channel="#forest-notifications", text=f"Actor mirroring failed: {e}"
>     ).validate()
> ```
> 
> We can chat about dependency management if you don't have a solution already
> 
> (I've not tested the above because I don't have the credentials required)
> 
> Also, I don't think every script should have to worry about failure - can we have a `job` that runs `on: failure` to report failing runs automatically?

I haven't, i will play with python and come up with a solution or text you.

LesnyRumcajs (changes requested):
1. You introduced a new scripting language into the repository. There should be a linter in the CI for that language. I don't know Python, so I'd defer to such tools to check for language issues.

2. It'd be great to make it possible for the script to run locally and not necessarily uploading to S3 buckets. This could be achieved by abstracting the sink (in this case, `s3.put_object`).

3. Can be a follow-up issue, but it'd be great to have it containerized.

.github/workflows/mirror-builtin-actors.yml
```diff
@@ -0,0 +1,86 @@
...
+      - name: Install dependencies
+        working-directory: scripts/mirror-actors
+        run: poetry install --no-interaction --no-root
+
+      - name: Mirror Actors to DigitalOcean
+        working-directory: scripts/mirror-actors
+        if: github.ref == 'refs/heads/main' && (github.event_name == 'push' || github.event_name == 'workflow_dispatch')
+        run: |
+          source .venv/bin/activate
+          python3 mirror_actors/
```
aatifsyed, line 41: `poetry run python -m mirror_actors`

.github/workflows/mirror-builtin-actors.yml
```diff
@@ -0,0 +1,86 @@
...
+      - name: Install dependencies
+        working-directory: scripts/mirror-actors
+        run: poetry install --no-interaction --no-root
+
+      - name: Mirror Actors to CloudFlare
+        working-directory: scripts/mirror-actors
+        if: github.ref == 'refs/heads/main' && (github.event_name == 'push' || github.event_name == 'workflow_dispatch')
+        run: |
+          source .venv/bin/activate
+          python3 mirror_actors/
```
aatifsyed, line 78: As above

.github/workflows/scripts-lint.yml
```diff
@@ -38,3 +38,24 @@ jobs:
...
+      with:
+        virtualenvs-create: true
+
+    - name: Install dependencies
+      working-directory: scripts/mirror-actors
+      run: poetry install --no-interaction --no-root
+
+    - name: Lint Python Code
+      working-directory: scripts/mirror-actors
+      run: poetry run pylint mirror_actors/ -f actions
```
aatifsyed, line 61: Could also add black as a formatter, and pylance as a type checker if you really want "production grade python". 

Might not be relevant for quick scripts

Happy for this to be deferred as an issue

scripts/mirror-actors/README.md
```diff
@@ -0,0 +1,46 @@
+# Overview
```
aatifsyed, line 1: *excellent* readme

scripts/mirror-actors/pyproject.toml
```diff
@@ -0,0 +1,23 @@
...
+version = "0.1.0"
+description = ""
+authors = ["samuelarogbonlo <samuel@chainsafe.io>"]
+readme = "README.md"
+
+[tool.poetry.dependencies]
+python = "^3.12"
+requests = "^2.31.0"
+slack-sdk = "^3.26.2"
+boto3 = "^1.34.22"
```
aatifsyed, line 12: There's also `boto3-types[S3]` iirc which helps pylance. Might be worth adding as a dev dep

aatifsyed (approved):
I love it

ruseinov:
Looks good, however since we're running this via cron - why don't we also make sure that dependencies are cahched? 
[Here's a guide.](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows)

samuelarogbonlo:
> 1. You introduced a new scripting language into the repository. There should be a linter in the CI for that language. I don't know Python, so I'd defer to such tools to check for language issues.
> 2. It'd be great to make it possible for the script to run locally and not necessarily uploading to S3 buckets. This could be achieved by abstracting the sink (in this case, `s3.put_object`).
> 3. Can be a follow-up issue, but it'd be great to have it containerized.

@LesnyRumcajs done, see [issue](https://github.com/ChainSafe/forest-iac/issues/390)

.github/workflows/scripts-lint.yml
samuelarogbonlo: see [issue](https://github.com/ChainSafe/forest-iac/issues/390)

samuelarogbonlo:
> Looks good, however since we're running this via cron - why don't we also make sure that dependencies are cahched? [Here's a guide.](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows)

@ruseinov done

ruseinov:
> > Looks good, however since we're running this via cron - why don't we also make sure that dependencies are cahched? [Here's a guide.](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows)
> 
> @ruseinov done

How about caching the installation itself: https://github.com/snok/install-poetry?tab=readme-ov-file#caching-the-poetry-installation

There are also useful guides on how to cache deps in the above readme too.

samuelarogbonlo:
> https://github.com/snok/install-poetry?tab=readme-ov-file#caching-the-poetry-installation

@ruseinov done

ruseinov (approved):

LesnyRumcajs:
@samuelarogbonlo I appreciate the effort of making it enterprise-grade, but a simple lambda would do enough for having a configurable sink. We're getting into non-trivial Python code where we shouldn't.

Pseudocode of what I mean:
```
sink = if use_local
  || save_to_file(file)
else
  || save_to_s3(file)
end
```

You also added quite a bit of unused methods. Let's keep things as simple as possible.

LesnyRumcajs:
If you guys are okay with that, then I won't block it.

samuelarogbonlo:
> If you guys are okay with that, then I won't block it.

@LesnyRumcajs. Good?

scripts/mirror-actors/mirror_actors/__main__.py
```diff
@@ -0,0 +1,103 @@
...
+    Send an alert message to a predefined Slack channel.
+    """
+    slack_api_token = os.environ["SLACK_API_TOKEN"]
+    slack_channel = os.environ["SLACK_CHANNEL"]
+    slack = WebClient(token=slack_api_token)
+    slack.chat_postMessage(channel=slack_channel, text=message).validate()
+
+def list_s3_objects(bucket_name, endpoint_url, region_name):
+    """
+    List all objects in an S3 bucket.
```
LesnyRumcajs, line 35: Is the documentation really needed here? The method names are self-describing.

LesnyRumcajs (approved):

scripts/mirror-actors/mirror_actors/__main__.py
samuelarogbonlo: This done to make the Pylint Happy

ruseinov (approved):

ruseinov:
> If you guys are okay with that, then I won't block it.

I kinda get where you are coming from, but I don't see a better way here. This is not too advances, just using some OOP that happens to be quite ugly in Python :)

PR #358, opened 2023-12-11
CF: redirect to archive when requesting latest snapshot

lemmih:
This lets people resume downloads even if new snapshots are uploaded.

**Summary of changes**
- `/latest/{chain}/` now redirects to `/archive/{chain}/latest/{specific_file.forest.car.zst}`

Closes

### DIFF
--- a/cf/latest-snapshot/src/index.ts
+++ b/cf/latest-snapshot/src/index.ts
@@ -36,7 +36,7 @@ async function get_latest(
 			// Response.redirect requires an absolute URL. Manually create the response to get around this.
 			return new Response('Found!', {
 				status: 302,
-				headers: { 'Location': '/archive/' + latest.key }
+				headers: { Location: '/archive/' + latest.key },
 			});
 		}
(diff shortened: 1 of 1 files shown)


elmattic (approved):

samuelarogbonlo (approved):

LesnyRumcajs (approved):

LesnyRumcajs:
@lemmih @ohmpatel1997 Is this going to be fine with regards to the latest snapshot metrics on Datadog?

lemmih:
> > @lemmih @ohmpatel1997 Is this going to be fine with regards to the latest snapshot metrics on Datadog?
> 
> this can possibly add inaccuracy to the metrics since if someone requested latest, we are redirecting him to archive, which is considered another request. but actually its only 1 request, for latest, instead of 2. i would suggest to not redirect but serve the same request.

Directly serving the request breaks resume functionality. Let's fix the logging by separating the `/archive/` worker and the `/latest/` worker.

PR #400, opened 2024-01-30
dockerize snapshot upload

LesnyRumcajs:
**Summary of changes**
Changes introduced in this pull request:
- dockerized snapshot service, which can now be run completely locally with only Docker as a requirement

I'll pin the snapshot version in the TF to a specific version once it is built by CI (it needs to reach `main` first).

**Reference issue to close (if applicable)**

Part of https://github.com/ChainSafe/forest-iac/issues/364

**Other information and links**

PR #403, opened 2024-02-09
Resurrect snapshot container

LesnyRumcajs:
**Summary of changes**
- resurrected [the](https://github.com/ChainSafe/forest-iac/pull/400) with few changes:
  - the base snapshot service image doesn't have the `aws-cli` dependency anymore - it is replaced by the usage of https://gallery.ecr.aws/aws-cli/aws-cli (used the ECR flavour to bypass Dockerhub download limits)
  - no copies are made, the volume is mounted directly to the `aws-cli` container which uses it to upload the snapshot.

The peak disk usage will be 2x snapshot size for a brief moment (the period of snapshot export command), but apart from that, e.g., during the upload, it should be 1x snapshot size.

Closes


ruseinov:
I'll be honest, the setup for snapshot service is looking pretty confusing, a mix of ruby, bash and bash embedded into ruby. Perhaps we should improve upon this code? My eyes bleed reading this, it's not comprehensive at all. And we are talking about a really small amount of code. 

If we want to keep using Ruby/Python or whatever else scripting language here - I think we should reconsider how `bash` is wrapped, and come up with somewhat clearer abstractions, e.g. an Object that has a bunch of methods that then do some `bash` stuff under the hood.  It would be much nicer than interpolating a bunch of commands into a single string. 

This stuff is looking pretty unreadable:
```
add_timestamps_cmd = <<~CMD
  awk '{
  if ($0 !~ /^[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}.[0-9]{6}Z/)
    print strftime("[%Y-%m-%d %H:%M:%S]"), $0;
  else
    print $0;
  fflush();
  }'
CMD
```
I appreciate the `awk` knowledge, but I'd prefer to have a readable alternative. If it's not easier and more comprehensive to do it in Ruby - at least some comments to avoid having to decipher this would be much appreciated.


All of that said, if it's too much effort to invest - it would at least be nice to have this code properly documented.

LesnyRumcajs:
> I'll be honest, the setup for snapshot service is looking pretty confusing, a mix of ruby, bash and bash embedded into ruby. Perhaps we should improve upon this code? My eyes bleed reading this, it's not comprehensive at all. And we are talking about a really small amount of code.
> 
> If we want to keep using Ruby/Python or whatever else scripting language here - I think we should reconsider how `bash` is wrapped, and come up with somewhat clearer abstractions, e.g. an Object that has a bunch of methods that then do some `bash` stuff under the hood. It would be much nicer than interpolating a bunch of commands into a single string.
> 
> This stuff is looking pretty unreadable:
> 
> ```
> add_timestamps_cmd = <<~CMD
>   awk '{
>   if ($0 !~ /^[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}.[0-9]{6}Z/)
>     print strftime("[%Y-%m-%d %H:%M:%S]"), $0;
>   else
>     print $0;
>   fflush();
>   }'
> CMD
> ```
> 
> I appreciate the `awk` knowledge, but I'd prefer to have a readable alternative. If it's not easier and more comprehensive to do it in Ruby - at least some comments to avoid having to decipher this would be much appreciated.
> 
> All of that said, if it's too much effort to invest - it would at least be nice to have this code properly documented.

@ruseinov The code to review is here https://github.com/ChainSafe/forest-iac/pull/403/commits/2f263d09da1a4495147623cd9ca701b5c3acc0d5. You don't have to review the `awk` monstrosity at all.

The overall design of the service could definitely be improved, but it's not the point of this PR. The point is to improve upon its usage, e.g., via Terraform.

lemmih (approved):

samuelarogbonlo (approved):
LGTM

PR #438, opened 2024-05-03
feat: allow specifying calibnet and/or mainnet snapshot for the daily…

### LINKED ISSUE #376: Allow specifying calibnet and/or mainnet snapshot for the daily snapshot
**Issue summary**

Right now, we spawn calibnet and mainnet services on a single machine. Potentially, we would like to have the `mainnet` snapshot generation isolated (or just generating calibnet snapshots for testing).

Make the `forest-snapshot` module configurable in a way that allows specifying whether we want it to run mainnet-only, calibnet-only or both.

samuelarogbonlo:
**Summary of changes**
- allow specifying calibnet and/or mainnet snapshot for the daily snapshot

Closes #376 


tf-managed/modules/daily-snapshot/service/init.sh
```diff
@@ -21,5 +21,18 @@ done'
 # Run new_relic and fail2ban scripts
 bash newrelic_fail2ban.sh
 
-# Setup cron jobs
-cp calibnet_cron_job mainnet_cron_job /etc/cron.hourly/
+# Setup cron job
+case "$SNAPSHOT_TYPE" in
+    mainnet)
+        cp mainnet_cron_job /etc/cron.hourly/ && rm calibnet_cron_job
```
LesnyRumcajs, line 27: isn't `cp` and `rm` a `mv`?

tf-managed/modules/daily-snapshot/variable.tf
```diff
@@ -109,6 +109,13 @@ variable "environment" {
   type        = string
 }
 
+
+variable "snapshot_type" {
```
LesnyRumcajs, line 113: It'd be more extensible (and readable) if we accept a set of networks. On top of that, if we restrict possible values via comment, we can as well restrict them via validations.

samuelarogbonlo: True, done.

tf-managed/modules/daily-snapshot/service/init.sh
samuelarogbonlo: yeah, skipped true.

tf-managed/modules/daily-snapshot/variable.tf
samuelarogbonlo: i've added the validations, but i'm abit confused about the `accept a set of network`.

samuelarogbonlo: ping @LesnyRumcajs

LesnyRumcajs:
```
 type    = set(string)
  default = ["mainnet", "calibnet"]
```
Then, in the script, you would iterate over those values and generate proper cron jobs.

LesnyRumcajs: This approach is more flexible than the one with `both`, as it allows to possibly add more values in the future. Say, we add `pineconet`. Now the allowed values would be `mainnet, calibnet, pineconenet, both`? Seems weird. We'd change `both` to `all` but that would be a breaking change. We should design, if possible, for future changes.

samuelarogbonlo: sure, done @LesnyRumcajs

LesnyRumcajs: `all` should not be a network name. The variable itself should be named `snapshot_networks` and, for the moment, it should accept either `mainnet`, `calibnet` or `mainnet,calibnet`. By default `mainnet,calibnet`

samuelarogbonlo: @LesnyRumcajs done

samuelarogbonlo:
ping @LesnyRumcajs

tf-managed/modules/daily-snapshot/variable.tf
```diff
@@ -109,6 +109,18 @@ variable "environment" {
...
+  type        = set(string)
+  description = "Set of network types for snapshot generation. Valid options: 'mainnet' and/or 'calibnet'"
+  default     = ["mainnet", "calibnet"]
+
+  validation {
+    condition     = alltrue([for val in var.snapshot_type : contains(["mainnet", "calibnet"], val)])
```
LesnyRumcajs, line 119: Can we also check that the list is not empty?

tf-managed/modules/daily-snapshot/service/init.sh
```diff
@@ -21,5 +21,20 @@ done'
...
+      mv mainnet_cron_job /etc/cron.hourly/
+      ;;
+    calibnet)
+      mv calibnet_cron_job /etc/cron.hourly/
+      ;;
+    *)
+      echo "Error: Invalid network type '$type'"
+      ;;
+  esac
+done
```
LesnyRumcajs, line 40: It can be a follow-up, but we can generate the cron scripts now, instead of copying pre-defined ones.

LesnyRumcajs (approved):

tf-managed/modules/daily-snapshot/variable.tf
samuelarogbonlo: @LesnyRumcajs. done, please re-approve.

tf-managed/modules/daily-snapshot/service/init.sh
samuelarogbonlo: sure sure.

LesnyRumcajs:
Forest snapshot service discontinued in favour of the infra snapshot service.

Issue #424, opened 2024-03-12
Timeout on remote exec


LesnyRumcajs:
**Issue summary**

While deploying node or snapshot services, Terraform seems to have had an issue with SSH connectivity. For example, [this](https://github.com/ChainSafe/forest-iac/actions/runs/8244374952/job/22549201743) succeeded on 3rd attempt; previous ones reported a timeout.

```
digitalocean_droplet.forest (remote-exec): Connecting to remote host via SSH...
digitalocean_droplet.forest (remote-exec):   Host: 209.38.234.101
digitalocean_droplet.forest (remote-exec):   User: root
digitalocean_droplet.forest (remote-exec):   Password: false
digitalocean_droplet.forest (remote-exec):   Private key: false
digitalocean_droplet.forest (remote-exec):   Certificate: false
digitalocean_droplet.forest (remote-exec):   SSH Agent: true
digitalocean_droplet.forest (remote-exec):   Checking Host Key: false
digitalocean_droplet.forest (remote-exec):   Target Platform: unix
digitalocean_droplet.forest: Still creating... [5m0s elapsed]
digitalocean_droplet.forest: Still creating... [5m10s elapsed]
digitalocean_droplet.forest: Still creating... [5m20s elapsed]
digitalocean_droplet.forest (remote-exec): Connecting to remote host via SSH...
digitalocean_droplet.forest (remote-exec):   Host: 209.38.234.101
digitalocean_droplet.forest (remote-exec):   User: root
digitalocean_droplet.forest (remote-exec):   Password: false
digitalocean_droplet.forest (remote-exec):   Private key: false
digitalocean_droplet.forest (remote-exec):   Certificate: false
digitalocean_droplet.forest (remote-exec):   SSH Agent: true
digitalocean_droplet.forest (remote-exec):   Checking Host Key: false
digitalocean_droplet.forest (remote-exec):   Target Platform: unix
digitalocean_droplet.forest: Still creating... [5m30s elapsed]
digitalocean_droplet.forest: Still creating... [5m40s elapsed]
digitalocean_droplet.forest: Still creating... [5m50s elapsed]
digitalocean_droplet.forest: Still creating... [6m0s elapsed]
digitalocean_droplet.forest (remote-exec): Connecting to remote host via SSH...
digitalocean_droplet.forest (remote-exec):   Host: 209.38.234.101
digitalocean_droplet.forest (remote-exec):   User: root
digitalocean_droplet.forest (remote-exec):   Password: false
[...]

PR #439, opened 2024-05-03
fix: increase remote exec timeout

samuelarogbonlo:
**Summary of changes**
- increase remote exec timeout

part of #424 


hanabi1224 (approved):

LesnyRumcajs (changes requested):
It's not only the daily snapshot that fails on timeouts; other services should be modified as well.

samuelarogbonlo:
> It's not only the daily snapshot that fails on timeouts; other services should be modified as well.

done.

LesnyRumcajs:
What about `forest-droplet`?

lemmih (approved):

Issue #370, opened 2024-01-09
Hosted snapshots should include their respective hashes


LesnyRumcajs:
**Issue summary**

This is an umbrella issue for all snapshots hosted in Cloudflare. If someone downloads 20 TiB worth of snapshots, they should be able to verify successful and correct download via digests.

This includes:
- [ ] Lite snapshots (and their automatic generation, once it happens),
- [ ] Diff snapshots (and their automatic generation, once it happens),
- [x] Latest snapshots

For both mainnet and calibration network.

The standard approach is to use `sha256sum`. Unless something has changed, Forest can generate the checksum file independently when exporting a snapshot. I'm not sure what the status of this feature outside of the export command is.

**Other information and links**

samuelarogbonlo:
@LesnyRumcajs, my proposed solution is to write a one off script will download or regenerate the current snapshots (both lite and diff), create a shasum file for them, and upload everything to R2. After that,  we can update [this](https://github.com/ChainSafe/forest-iac/pull/428) when @sudo-shashank is done to include the SHA sum in the process of generating and uploading archival snapshots. For the latest snapshot, we can update the snapshot service to also generate shasums and upload. 

Question: should we update the Cf code to also download the `shasum` when the snapshot link is hit

LesnyRumcajs:
1. It's not needed to download the lite and diff snapshots (at least most of them) - they are in majority on archie.
2. What's the point of regenerating the snapshots? The snapshots won't change, so only corresponding shasum files need to be uploaded.

> Question: should we update the Cf code to also download the shasum when the snapshot link is hit
No. The vast majority of people won't use them.

LesnyRumcajs:
Migrated to Infrastructure repo.

PR #441, opened 2024-05-08
fix: upload corresponding latest snapshot sha256sum 

samuelarogbonlo:
**Summary of changes**
- upload corresponding latest snapshot sha256sum 

part #370 


lemmih (changes requested):
Let's enumerate the files to be uploaded rather than using a wildcard.

samuelarogbonlo:
> Let's enumerate the files to be uploaded rather than using a wildcard.

updated, @lemmih what do you think ?

LesnyRumcajs:
@lemmih Looking at https://github.com/ChainSafe/forest-iac/blob/main/cf/prune-latest/src/index.ts, it seems that adding corresponding checksum files would decrease the number of recent snapshots by half. Are we okay with that? Or should that worker be updated accordingly?

lemmih:
> @lemmih Looking at https://github.com/ChainSafe/forest-iac/blob/main/cf/prune-latest/src/index.ts, it seems that adding corresponding checksum files would decrease the number of recent snapshots by half. Are we okay with that? Or should that worker be updated accordingly?

If the Infra team is ready to take over then we don't have to worry about retention. Let's see what Josh at the meeting later today. If they need more time, we'll bump our retention (which is still quite a bit lower than required by the contract, iirc).

hanabi1224 (approved):

LesnyRumcajs:
Service snapshot discontinued in favour of infra snapshot service.

PR #444, opened 2024-05-10
fix: add retry step to terragrunt apply

samuelarogbonlo:
**Summary of changes**
- add retry step to terragrunt apply

Partof #424 

### DIFF
--- a/composite-action/terragrunt/action.yml
+++ b/composite-action/terragrunt/action.yml
@@ -33,6 +33,12 @@ inputs:
     description: 'CloudFlare R2 access key id'
   r2_secret_key:
     description: 'CloudFlare R2 private access key'
+  max_retries:
+    description: 'The maximum number of retry attempts for the Terragrunt apply command if it fails initially.'
+    required: true
+  retry_delay_seconds:
+    description: 'The number of seconds to wait between each retry attempt.'
+    required: true
 
 runs:
   using: "composite"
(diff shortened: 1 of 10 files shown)


tf-managed/modules/daily-snapshot/main.tf
```diff
@@ -75,10 +75,9 @@ resource "digitalocean_droplet" "forest" {
   graceful_shutdown = false
 
   connection {
-    host    = self.ipv4_address
-    user    = "root"
-    type    = "ssh"
-    timeout = "30m"
```
ruseinov, line 81: why the timeout change?

aatifsyed (approved):

tf-managed/modules/daily-snapshot/main.tf
LesnyRumcajs: Fail fast and retry. If it doesn't finish successfully in a few minutes, given how trivial the initialisation is, it will likely not succeed in 30 minutes as well.

composite-action/terragrunt/action.yml
```diff
@@ -155,7 +161,22 @@ runs:
         else
           echo "Changes detected. Redeploying everything..."
           terragrunt destroy -auto-approve --terragrunt-non-interactive
```
LesnyRumcajs, line 163: I believe this is the issue that should be addressed. We should use `lifecycle.create_before_destroy` and not force-destroy the resource - it should get destroyed after the new one is created successfully, leaving it untouched in case of errors. What do you think? We can do it in a separate issue/PR.

ruseinov (approved):

composite-action/terragrunt/action.yml
samuelarogbonlo: yeah makes sense. i will implement in this PR

LesnyRumcajs:
@samuelarogbonlo  The `lifecycle.create_before_destroy` won't work if we do `terragrunt destroy -auto-approve --terragrunt-non-interactive`, no?

samuelarogbonlo:
> @samuelarogbonlo The `lifecycle.create_before_destroy` won't work if we do `terragrunt destroy -auto-approve --terragrunt-non-interactive`, no?

@LesnyRumcajs, `lifecycle.create_before_destroy` is useless in this current set-up will remove.

LesnyRumcajs:
Given we only have sync check now, it doesn't seem required.

PR #529, opened 2026-08-10
feat: build & publish `rpc-checks` docker image

EclesioMeloJunior:
**Summary of changes**

1. Remove `.except('logsBloom')` from `check_rpc.rb`
2. Run image on `ruby:4.0-alpine` to reduce final image size
3. Add `.github/workflows/rpc-checks-image.yml` that builds & publish the image to `GHCR`

Ref: https://github.com/ChainSafe/forest/issues/7270

N/A

### DIFF
--- a/.github/workflows/rpc-checks-image.yml
+++ b/.github/workflows/rpc-checks-image.yml
@@ -1,11 +1,9 @@
 name: rpc-checks image
 
 on:
-  pull_request:
-    paths:
-      - 'scripts/rpc-checks/**'
-      - '.github/workflows/rpc-checks-image.yml'
   push:
+    branches:
+      - main
     paths:
       - 'scripts/rpc-checks/**'
       - '.github/workflows/rpc-checks-image.yml'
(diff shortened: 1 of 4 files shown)


.github/workflows/rpc-checks-image.yml
```diff
@@ -0,0 +1,43 @@
...
+    branches:
+      - main
+    paths:
+      - 'scripts/rpc-checks/**'
+      - '.github/workflows/rpc-checks-image.yml'
+  workflow_dispatch:
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
```
LesnyRumcajs: nit: this could be cheap (just packaging a ruby script) enough to run on ubuntu slim

scripts/rpc-checks/check_rpc.rb
```diff
@@ -46,18 +46,14 @@
 
 def hex(num) = format('0x%x', num)
 
-# Known expected mismatches, dropped before comparing (remove as they get
```
LesnyRumcajs, line 49: I don't get it, you removed the bloom filter suppression (great), but also removed the doc for `accessList` which is a separate issue, without removing the code. You either need to bring back the doc for this or remove the workaround if it's not needed anymore.

LesnyRumcajs (changes requested):

LesnyRumcajs (approved):

